PR #2, opened 2015-02-10
Fix image upload when using G+-Photos app

leonhandreke:
When using the G+-Photos app, it seems that photo URIs returned by the pick image intent are no longer associated with a filename. Use the activity's content resolver to get a stream instead and use that for uploading.

I have not tested this with the AOSP Gallery app because my phone (Nexus 4 with factory L image) only has the G+-Photos app. It would be great if somebody with this app could try the patch.

`ContentManager` now no longer sets `Content-Length`. During my testing on matrix.org, this didn't seem to be a problem, but reading the synapse source it actually should (see https://github.com/matrix-org/synapse/blob/771892b3144e5a45739a097f818e71a7ce5d8ce2/synapse/media/v1/upload_resource.py#L48). Not sure what's going on here.

### DIFF
--- a/app/src/main/java/org/matrix/matrixandroidsdk/activity/RoomActivity.java
+++ b/app/src/main/java/org/matrix/matrixandroidsdk/activity/RoomActivity.java
@@ -5,9 +5,8 @@ import android.app.ProgressDialog;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.net.Uri;
-import android.support.v4.app.FragmentManager;
-import android.support.v7.app.ActionBarActivity;
 import android.os.Bundle;
+import android.support.v4.app.FragmentManager;
 import android.text.TextUtils;
 import android.util.Log;
 import android.view.Menu;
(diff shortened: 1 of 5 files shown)


Kegsay:
This looks good (inspected diff only; not tested), but I would prefer that the logic to go from the ``Intent`` to the ``InputStream`` be factored out into ``ResourceUtils`` or another appropriate utility class, to avoid the duplication that has sprouted here (``RoomActivity:312`` and ``SettingsActivity:227``)

In addition, as bad form as it is, please wrap your calls to ``openInputStream`` and ``getType`` in a generic ``Exception`` block, not just ``FileNotFoundException``. This is because some versions of Android will throw undocumented exceptions like ``IllegalStateException`` when accessing these APIs. By not catching them, this will crash the app, but they are entirely recoverable errors.

leonhandreke:
Thanks for the hint about Android API nastyness regarding exceptions. I've added a blanket catch-all.

I've moved the code for opening the resources to `ResourceUtil` as you suggested. I used `null` as a catch-all error return value to make the code a bit simpler for callers. (Is there something like `Optional` in Guava for Android?)

Kegsay:
Fixes look good, thanks! I'm not sure on ``Optional`` for Guava. Will give it until tomorrow for anyone else to chime in, then merge this in.

PR #4, opened 2015-02-12
Size thumbnail in image message dependent on the size of the view

leonhandreke:
This makes thumbnail display resolution-independent.


Kegsay:
LGTM

PR #5, opened 2015-02-15
Don't bing if no rules match

leonhandreke:
The default rules in synapse are set up to match and notify when
mentioned in a group chat and always in 1-to-1 conversations. To have
sane defaults, don't notify in all other cases.

### DIFF
--- a/sdk/src/main/java/org/matrix/androidsdk/util/BingRulesManager.java
+++ b/sdk/src/main/java/org/matrix/androidsdk/util/BingRulesManager.java
@@ -107,8 +107,8 @@ public class BingRulesManager {
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


Kegsay:
Just checked with Dave; the default is to bing if no rules match, as per:
 - https://github.com/matrix-org/synapse/blob/master/synapse/push/__init__.py#L130
 - https://github.com/matrix-org/synapse/blob/master/synapse/push/__init__.py#L35

so I can't accept this PR, sorry!

ara4n:
The correct solution is to change the default in the server and/or make it explicit and disablable - not work around the thinko in the client. We can and will do this tho.

PR #6, opened 2015-02-23
Add pushers API client and GCM skeleton

leonhandreke:


app/src/main/AndroidManifest.xml
```diff
@@ -9,6 +9,15 @@
...
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
+    <!-- Google Cloud Messaging -->
+    <uses-permission android:name="com.google.android.c2dm.permission.RECEIVE" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <permission
+        android:name="${applicationId}.permission.C2D_MESSAGE"
+        android:protectionLevel="signature" />
+    <uses-permission
+        android:name="${applicationId}.permission.C2D_MESSAGE" />
```
Kegsay: Where is ``applicationId`` coming from in this case? I haven't seen this used before.

Kegsay:
LGTM other than the query about ``${applicationId}``.

leonhandreke:
Re: `${applicationId}`: I saw it in some example around the internet and a quick search confirmed that it should work (and it does work, I've already successfully delivered GCM messages to the app).

http://tools.android.com/tech-docs/new-build-system/user-guide/manifest-merger seems to indicate that it uses the `applicationId` variable from the build.gradle.

Kegsay:
Awesome, I didn't know that. I'm more familiar with ANT when it comes to building Android projects, so this is nice to know.

PR #9, opened 2015-02-28
removing IDE specific files

SkaveRat:
.idea and iml files are local files, specific to the workspace and IDE settings. Checking these in will only dirty up your git environment


SkaveRat:
**project specific settings will be lost.**

*please note*: this will remove the files locally. best way is to close Android Studio before pulling and opening it up again as "open existing project"

not tested, but this will very likely work: move `.idea` and all `iml` files somewhere and move them back afterwards to keep project settings

Kegsay:
I'm of two minds with this one. The ``.idea`` files in question are correctly using ``$VARIABLES`` so there isn't any hard-coded conflicts which can arise from keeping them. Keeping these files means opening the project in Android Studio "Just Works" from a fresh ``git clone``. Of course, this in practice doesn't always happen and Android Studio does like to randomly clobber some of these files and/or remove the ``$VARIABLES`` being used, so I can see why you want to just remove them. 

I'm going to leave this open to allow other Android devs to give feedback, but if there isn't any major feeling either way, I'll accept this, since I feel the slightly higher cost of the initial setup is worth it to remove the pain of misbehaving IDEs.

SkaveRat:
> Keeping these files means opening the project in Android Studio "Just Works" from a fresh git clone

it still works perfectly fine. try cloning the branch and opening it.

Just the act of opening the project and creating my other PR caused every iml file and the .idea diretory being toched and diff'd. messy to commit nicely

Kegsay:
*nods* I'll give it a go and let you know how it goes.

PR #10, opened 2015-02-28
adding send intent receiving

SkaveRat:
Now possible to send text (URLs, tweets etc) via "share" in other apps to matrix rooms

(dupe of messed up PR #8 )

### DIFF
--- a/app/src/main/java/org/matrix/matrixandroidsdk/activity/HomeActivity.java
+++ b/app/src/main/java/org/matrix/matrixandroidsdk/activity/HomeActivity.java
@@ -27,6 +27,7 @@ import android.view.View;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 import android.widget.ExpandableListView;
+import android.widget.TextView;
 
 import org.matrix.androidsdk.MXSession;
 import org.matrix.androidsdk.data.Room;
@@ -293,6 +294,15 @@ public class HomeActivity extends MXCActionBarActivity {
                 if (null != roomId) {
                     CommonActivityUtils.goToRoomPage(roomId, HomeActivity.this);
                 }
+                if (Intent.ACTION_SEND.equals(getIntent().getAction()) && getIntent().getType() != null) {
+                    if ("text/plain".equals(getIntent().getType())) {
+                        CommonActivityUtils.goToRoomPage(roomId, HomeActivity.this, getIntent().getStringExtra(Intent.EXTRA_TEXT));
+                    }
+                }else {
+                    CommonActivityUtils.goToRoomPage(roomId, HomeActivity.this);
+                }
+                mAdapter.notifyDataSetChanged();
+
                 return true;
             }
         });
--- a/app/src/main/java/org/matrix/matrixandroidsdk/activity/RoomActivity.java
+++ b/app/src/main/java/org/matrix/matrixandroidsdk/activity/RoomActivity.java
@@ -71,7 +71,8 @@ import java.util.TimerTask;
  */
 public class RoomActivity extends MXCActionBarActivity implements MatrixMessageListFragment.MatrixMessageListFragmentListener {
 
-    public static final String EXTRA_ROOM_ID = "org.matrix.matrixandroidsdk.RoomActivity.EXTRA_ROOM_ID";
+    public static final String EXTRA_ROOM_ID         = "org.matrix.matrixandroidsdk.RoomActivity.EXTRA_ROOM_ID";
+    public static final String EXTRA_INITIAL_MESSAGE = "org.matrix.matrixandroidsdk.RoomActivity.EXTRA_INITIAL_MESSAGE";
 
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/matrix/matrixandroidsdk/activity/RoomActivity.java
```diff
@@ -61,7 +61,8 @@
  */
 public class RoomActivity extends MXCActionBarActivity implements MatrixMessageListFragment.MatrixMessageListFragmentListener {
 
-    public static final String EXTRA_ROOM_ID = "org.matrix.matrixandroidsdk.RoomActivity.EXTRA_ROOM_ID";
+    public static final String EXTRA_ROOM_ID         = "org.matrix.matrixandroidsdk.RoomActivity.EXTRA_ROOM_ID";
```
Kegsay: Please keep to the style of other constants in this file. Do not align the value with other constant values.

The main reason for discouraging this is to avoid the maintenance burden of having to align all your constant values (which may be trivial initially, but quickly becomes annoying when you want to change constant names, etc).

app/src/main/java/org/matrix/matrixandroidsdk/activity/HomeActivity.java
```diff
@@ -257,7 +259,13 @@ public boolean onChildClick(ExpandableListView parent, View v,
                     roomId = mAdapter.getPublicRoomAt(childPosition).roomId;
                 }
 
-                CommonActivityUtils.goToRoomPage(roomId, HomeActivity.this);
+                if (Intent.ACTION_SEND.equals(getIntent().getAction()) && getIntent().getType() != null) {
```
Kegsay: Parsing the ``Intent`` here seems wrong to me. From this, it looks like if I were to share something with Matrix, I would be dumped into the Home Screen without any indication that I'm sharing something, then only when I click on a room will it auto-populate the text field with the text I planned on sharing. 

It may be worth having another ``Activity`` manage external ``Intents`` like this, so you can present a screen that much more clearly says "Select a room to share [text/image/whatever]". This would have the added benefit of keeping ``HomeActivity``'s complexity down, as it would be handling less things.

SkaveRat: Made a standalone activity at first, but noticed that the room-listing would need a refactor for this at first, as currently it's pretty hardwired to the `HomeActivity`.

This was a way to at least be able to share links, and refactor the list-creation in a new PR, as it's a bigger task

Kegsay: Having thought about it a bit more, making another ``Activity`` may end up being a lot of work for this patch to do. I'd be happy with some visual indication that I was sharing something, and that I need to select a room to complete the action. Please also remember to kill the ``Intent`` (e.g. using ``activity.setIntent(null)``) after you have processed it, to avoid replaying it under certain ``Activity`` lifecycle changes.

SkaveRat:
added a "Share with:" title when coming from an Send-Intent.

I'm not convinced about setIntent(null), as this would make for a confusing UX.
Currently, you can send someone a link, hit back, select the next user and share the link again, as you basicly are still in the sharing-flow (hitting back again, would bring you back to the original app).

Kegsay:
> Currently, you can send someone a link, hit back, select the next user and share the link again, as you basicly are still in the sharing-flow

Seems reasonable.

Tested this; there's a few issues with this currently:
 - Sharing with Matrix from cold (app not open) dumps you into an empty home screen. I would expect this to be the splash screen then transfer you over to a non-empty home screen.
 - The share fork with the "share with" patch ( https://github.com/SkaveRat/matrix-android-sdk/tree/patch-sendintent ) isn't compiling (``alpha/debug/layout/activity_home.xml:21: error: Error: No resource found that matches the given name (at 'text' with value '@string/share_title'``) and sure enough a grep isn't turning up ``share_title``.

SkaveRat:
Hmm. not sure how to fix the coldboot problem. Looks like we need the new activity for sharing, which can detect a missig session, and log in first

Kegsay:
Yeah, it would be a decent amount of work to fix that, but we should be handling it better than that. It may be preferable to just ``finish()`` the the Activity if it isn't ready and was launched from the SEND Intent?

Kegsay:
Any further progress on this? If we can add in the ``finish()`` to avoid being dropped into a broken Home screen, I'm happy to merge this in.

SkaveRat:
sorry, didn't have much time.

the current code isn't working anymore, the text doesn't pre-fill the inpt anymore. Merging the current master seems to have broken it.

I'll check out the bug in the next days, and rebase the commits. It's a mess, currently

SkaveRat:
rebased. still doesn't work and can't seem to find the reason why. Text just doesn't get set.

Kegsay:
Going to close this for now. I really would like this feature, but there are too many edge cases hanging around due to the SDK's current code structure (tied too heavily to the Activity). The structure of the SDK needs to realistically be fixed first before this can be added.

PR #14, opened 2015-03-20
Add room name to notification text

ndarilek:


ndarilek:
A bit of additional info on this PR:

I recognize that it may be redundant to include the room name in multiple places. Unfortunately, it seems that Android AccessibilityServices only get the body of notifications, not their titles. So when my screen reader reads an incoming message, even though the room ID is displayed in the title, it only reads the message body because that is all that the ticker includes.

I added the room ID everywhere. If this takes up too much room, one possible alternative would be removing it from the body like it was before, replacing the ticker with a toast, and displaying the room ID + message in the toast only. Then it would take up less screen real-estate while still providing me speech feedback about which room originated a message.

Of course, this could entirely be a non-issue, and maybe it doesn't matter. Just thought I'd float an alternative if the notifications suddenly looked cramped. :)

Thanks.

ndarilek:
Actually, don't merge this yet, it didn't in fact add the room name/alias as I'd thought. This PR now adds the room ID, but of course that isn't what I want either.

Using the Android SDK, how do I resolve a room ID to its list of aliases? I'm assuming there's some sort of state-tracker that intercepts state events on rooms and maps their IDs to received values but maybe I'm wrong.

ndarilek:
Figured it out. I think this is ready to review.

app/src/main/java/org/matrix/matrixandroidsdk/services/EventStreamService.java
```diff
@@ -264,14 +266,20 @@ private void startWithNotification() {
         mState = StreamAction.START;
     }
 
-    private Notification buildMessageNotification(String from, String body, String roomId) {
+    private Notification buildMessageNotification(String from, String body, String roomId, String roomName) {
         NotificationCompat.Builder builder = new NotificationCompat.Builder(this);
         builder.setWhen(System.currentTimeMillis());
         builder.setContentTitle(from);
+        String name = null;
+        if("" == roomName || null == roomName) {
```
Kegsay: You can use ``TextUtils.isEmpty(roomName)`` here from ``android.text``, which conveys your intent more clearly.

app/src/main/java/org/matrix/matrixandroidsdk/services/EventStreamService.java
```diff
@@ -150,9 +150,11 @@ public void onBingEvent(Event event, RoomState roomState, BingRule bingRule) {
                 return;
             }
 
+            String roomName = room.getName(mSession.getMyUser().userId);
```
Kegsay: Is there ever a chance of ``mSession.getMyUser()`` being ``null`` (e.g. because they logged out and the notification arrives after they logout)? Not sure if this can happen; just asking if you've thought of this.

app/src/main/java/org/matrix/matrixandroidsdk/services/EventStreamService.java
```diff
@@ -264,14 +266,20 @@ private void startWithNotification() {
...
+        if("" == roomName || null == roomName) {
+            name = "";
+        } else {
+            name = " (" + roomName + "): ";
+        }
         builder.setContentText(body);
         builder.setAutoCancel(true);
         builder.setSmallIcon(R.drawable.ic_menu_small_matrix);
-        builder.setTicker(from + ":" + body);
+        builder.setTicker(from + name + body);
```
Kegsay: If ``name`` is not set, we're losing the ``:`` which was here before, as ``:275`` sets it to ``""``.

Kegsay:
Otherwise LGTM. Is there any way to determine if a screen reader is being used in the app? If we get reports of it being too cramped, it would be nice to keep this behaviour if we know a screen reader is being used, but remove it if it isn't.

ndarilek:
This should address the concerns raised.

Note that I started integrating with the GCM code but didn't finish. If room_name is included in the bundle then it should be picked up, but if not then the code shouldn't fail. I couldn't figure out quickly where this bundle was packed for hand-off to GCM, and figured I'd rather submit what I have and advocate for its merge since develop is a fast-moving target and merge conflicts aren't fun. :)

Thanks.

Kegsay:
LGTM

PR #16, opened 2015-03-25
Add app-global GcmRegistrationManager to register for push services

leonhandreke:
Registration must be done both with the GCM server and with the matrix
home server, supplying a token acquired from the GCM server.

Registration is done in an AsyncTask, therefore the Matrix pushers API
can be called in a synchronous fashion without impacting responsiveness.

Registration of the pusher with the Matrix server is currently done on
every startup by SplashActivity but should eventually become a one-time
background task that is not tied to a specific activity.


app/src/main/java/org/matrix/matrixandroidsdk/gcm/RegisterGcmPusherTask.java
```diff
@@ -0,0 +1,22 @@
...
+    private Context mContext;
+
+    RegisterGcmPusherTask(Context context) {
+        mContext = context;
+    }
+
+    @Override
+    protected Void doInBackground(Void... voids) {
+
+        return null;
```
Kegsay: What is this class doing? If it's a stub, please add a ``TODO`` stating what needs to be done here.

Kegsay:
LGTM other than comments (which I seem to have done on a commit rather than the full diff, sorry!)

leonhandreke:
Thank you for the review! I've rebased this onto `develop`, squashed the fixup commits and pushed the result to https://github.com/leonhandreke/matrix-android-sdk/tree/gcm-registration-develop. It's probably easiest if you pull from there.

Kegsay:
That's awesome, thanks. Could you please open a PR for ``gcm-registration-develop``, just so we can be consistent with how PRs are being created/processed/accepted, then I'll merge it in.

PR #17, opened 2015-03-25
Add missing pieces to receive and display GCM notifications

leonhandreke:

PR #18, opened 2015-03-26
Add missing pieces to receive and display GCM notifications

leonhandreke:
Re-created because #17 had the wrong target branch.


Kegsay:
LGTM

PR #20, opened 2015-04-02
Basic support for Android Auto

friedger:
This PR adds support for Android Auto.

Notifcations that are created by the EventStreamService or GCM messages are extended for in-car use. 
The user can send a reply directly from the car UI.

To test this you need to use a car with Android Auto or install the Message Simulator (available with the SDK manager) on your device.

Remark: for in-car use (as well as for Android Wear) notifications should also be able to somehow handle/show the history in the room if several new messages arrive. Currently, there is only one notification for all rooms and all messages.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -124,7 +124,17 @@
             </intent-filter>
         </receiver>
 
+        <receiver android:name="org.matrix.matrixandroidsdk.car.CarBroadcastReceiver">
+            <intent-filter>
+                <action android:name="org.matrix.matrixandroidsdk.ACTION_MESSAGE_HEARD" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="org.matrix.matrixandroidsdk.ACTION_MESSAGE_REPLY" />
+            </intent-filter>
+        </receiver>
+
         <service android:name=".gcm.GcmIntentService" />
+
         <service android:name="org.matrix.matrixandroidsdk.services.EventStreamService" />
 
         <provider
@@ -132,6 +142,9 @@
             android:name="org.matrix.matrixandroidsdk.db.ConsoleContentProvider"
             android:exported="true" />
 
+        <meta-data
+          android:name="com.google.android.gms.car.application"
+          android:resource="@xml/automotive_app_desc" />
     </application>
 
 </manifest>
--- a/app/src/main/java/org/matrix/matrixandroidsdk/car/CarBroadcastReceiver.java
+++ b/app/src/main/java/org/matrix/matrixandroidsdk/car/CarBroadcastReceiver.java
@@ -1,0 +1,78 @@
+package org.matrix.matrixandroidsdk.car;
+
+import android.app.NotificationManager;
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+import android.support.v4.app.RemoteInput;
+import android.util.Log;
+
+import org.matrix.androidsdk.MXSession;
+import org.matrix.androidsdk.data.Room;
+import org.matrix.androidsdk.rest.callback.ApiCallback;
+import org.matrix.androidsdk.rest.model.Event;
+import org.matrix.androidsdk.rest.model.MatrixError;
+import org.matrix.androidsdk.rest.model.Message;
+import org.matrix.matrixandroidsdk.Matrix;
+import org.matrix.matrixandroidsdk.util.NotificationUtils;
+
+public class CarBroadcastReceiver extends BroadcastReceiver {
+    private static final String TAG = CarBroadcastReceiver.class.getSimpleName();
+
+    @Override
+    public void onReceive(final Context context, final Intent intent) {
+        String roomId = intent.getStringExtra(NotificationUtils.EXTRA_ROOM_ID);
+        if (NotificationUtils.ACTION_MESSAGE_HEARD.equals(intent.getAction())) {
+            NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+            notificationManager.cancelAll();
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/matrix/matrixandroidsdk/services/EventStreamService.java
```diff
@@ -148,7 +161,7 @@ public void onBingEvent(Event event, RoomState roomState, BingRule bingRule) {
             NotificationManager nm = (NotificationManager) EventStreamService.this.getSystemService(Context.NOTIFICATION_SERVICE);
             nm.cancelAll();
 
-            nm.notify(MSG_NOTIFICATION_ID, n);
+            nm.notify(event.roomId, MSG_NOTIFICATION_ID, n);
```
friedger: This creates one notification per room. Is this desired?

app/src/main/java/org/matrix/matrixandroidsdk/car/CarBroadcastReceiver.java
```diff
@@ -0,0 +1,78 @@
...
+                }
+
+                @Override
+                public void onMatrixError(final MatrixError e) {
+                    Log.d(TAG, "sending reply failed with matrix error " + e.errcode + " " + e.error);
+                }
+
+                @Override
+                public void onUnexpectedError(final Exception e) {
+                    Log.d(TAG, "sending reply failed!", e);
```
Kegsay: It would be nice to log these at ``error`` level, instead of ``debug``. If this is a common failure case then ``debug`` is fine but you should add a comment explaining why it's a common failure case.

app/src/main/java/org/matrix/matrixandroidsdk/car/CarBroadcastReceiver.java
```diff
@@ -0,0 +1,78 @@
...
+    public void onReceive(final Context context, final Intent intent) {
+        String roomId = intent.getStringExtra(NotificationUtils.EXTRA_ROOM_ID);
+        if (NotificationUtils.ACTION_MESSAGE_HEARD.equals(intent.getAction())) {
+            NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+            notificationManager.cancelAll();
+        } else if (NotificationUtils.ACTION_MESSAGE_REPLY.equals(intent.getAction())) {
+            final CharSequence reply = getMessageText(intent);
+
+            final MXSession session = Matrix.getInstance(context).getDefaultSession();
+            final Room room = session.getDataHandler().getRoom(roomId);
```
Kegsay: ``getDefaultSession()`` can return ``null`` according to the docs at https://github.com/matrix-org/matrix-android-sdk/blob/master/app/src/main/java/org/matrix/matrixandroidsdk/Matrix.java#L43 - in which case the app will ``NullPointerException`` here.

app/src/main/java/org/matrix/matrixandroidsdk/util/NotificationUtils.java
```diff
@@ -19,6 +20,10 @@
 
     public static final String QUICK_LAUNCH_ACTION = "org.matrix.matrixandroidsdk.services.EventStreamService.QUICK_LAUNCH_ACTION";
     public static final String TAP_TO_VIEW_ACTION = "org.matrix.matrixandroidsdk.services.EventStreamService.TAP_TO_VIEW_ACTION";
+    public static final String CAR_VOICE_REPLY_KEY = "org.matrix.matrixandroidsdk.services.EventStreamService.CAR_VOICE_REPLY_KEY" ;
+    public static final String ACTION_MESSAGE_HEARD = "org.matrix.matrixandroidsdk.ACTION_MESSAGE_HEARD";
+    public static final String ACTION_MESSAGE_REPLY = "org.matrix.matrixandroidsdk.ACTION_MESSAGE_REPLY";
+    public static final String EXTRA_ROOM_ID = "org.matrix.matrixandroidsdk.EXTRA_ROOME_ID";
```
Kegsay: Nitpick: Typo in the constant value, "ROOME"

app/src/main/AndroidManifest.xml
```diff
@@ -124,7 +124,17 @@
             </intent-filter>
         </receiver>
 
+        <receiver android:name="org.matrix.matrixandroidsdk.car.CarBroadcastReceiver">
+            <intent-filter>
+                <action android:name="org.matrix.matrixandroidsdk.ACTION_MESSAGE_HEARD" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="org.matrix.matrixandroidsdk.ACTION_MESSAGE_REPLY" />
+            </intent-filter>
```
Kegsay: Nitpick: Both of these actions can be contained in a single ``<intent-filter>``, and in this case would make your intention clearer.

Kegsay:
LGTM other than comments.

app/src/main/java/org/matrix/matrixandroidsdk/car/CarBroadcastReceiver.java
friedger: The user should get some feedback, and the matrix error should be handled with maybe a resend?

I will investigate how to inform the user for e.g. network failure

app/src/main/java/org/matrix/matrixandroidsdk/car/CarBroadcastReceiver.java
friedger: This part is only reached if the user has already received a message, so the session should be not null. However, to be on the save side I'll add a case for null, i.e. the user clears the app's data while a notification is shown in the car's dashboard.

friedger:
@Kegsay thanks for the review.

friedger:
Just for documentation a screenshot that shows how Android Auto looks like (in this cause with one notification per room)

screenshot_2015-04-02-10-02-51

PR #22, opened 2015-04-23
Upgrade build tools version

ndarilek:


Kegsay:
Closing this issue because there isn't any reason to upgrade the build tools currently. If you feel that this is a mistake and we really really should be using `22.0.1` or above, please say *why* (e.g. if you had problems building the SDK) and we can bump the version accordingly.

PR #25, opened 2015-09-22
Basic impl of specifying cert fingerprints.

erikjohnston:
This PR aims to add the necessary SDK hooks to make it possible and easy for apps to handle self signed certs (by using cert finger prints and "trust on first use" ). The PR also contains some refactoring of the SDK APIs to make the necessary changes nicer and more consistent.


matrix-sdk/src/main/java/org/matrix/androidsdk/HomeserverConnectionConfig.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public HomeserverConnectionConfig(Uri hsUri) {
+        this(hsUri, null, null, false);
+    }
+
+    /**
+     * @param hsUri The URI to use to connect to the homeserver
+     * @param credentials The credentials to use, if needed. Can be null.
+     */
+    public HomeserverConnectionConfig(Uri hsUri,Credentials credentials) {
```
Kegsay: Nit: Missing space after `,`

matrix-sdk/src/main/java/org/matrix/androidsdk/HomeserverConnectionConfig.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public Uri getHomeserverUri() { return mHsUri; }
+    public byte[][] getAllowedFingerprints() { return mAllowedFingerprints; }
+
+    public Credentials getCredentials() { return mCredentials; }
+    public void setCredentials(Credentials credentials) { this.mCredentials = credentials; }
+
+    public boolean getPinned() {
+        return mPin;
```
Kegsay: Not intuitive what this is used for.

matrix-sdk/src/main/java/org/matrix/androidsdk/HomeserverConnectionConfig.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.util.ArrayList;
+
+
+/**
+ * Represents how to connect to a specific Homeserver, may include credentials to use.
+ */
+public class HomeserverConnectionConfig {
+    private Uri mHsUri;
+    private byte[][] mAllowedFingerprints;
```
Kegsay: I think we should probably yank the `byte[]` for the `fingerprint` to a `Fingerprint` class, which could also potentially store useful information like the hex-displayable string and when you accepted/rejected it. Also, do we ever anticipate adding to this array? Would an `ArrayList` be better?

matrix-sdk/src/main/java/org/matrix/androidsdk/ssl/CertUtil.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param cert
+     * @return
+     */
+    public static byte[] generateSha256Fingerprint(X509Certificate cert) {
+        final byte[] fingerprint;
+        try {
+            MessageDigest md = MessageDigest.getInstance("SHA256");
+            fingerprint = md.digest(cert.getEncoded());
+        } catch(Exception e) {
+            throw new RuntimeException(e);
```
Kegsay: Is a `RuntimeException` really appropriate here? This isn't recoverable in any way? (e.g. by rejecting the cert?)

matrix-sdk/src/main/java/org/matrix/androidsdk/ssl/PinnedTrustManager.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public void checkClientTrusted(X509Certificate[] chain, String s) throws CertificateException {
+        try {
+            if (mDefaultTrustManager != null) {
+                mDefaultTrustManager.checkClientTrusted(
+                        chain, s
+                );
+            }
+        } catch (CertificateException e) {
+            // If there is an exception we full back to checking fingerprints
```
Kegsay: `s/full/fall/`

matrix-sdk/src/main/java/org/matrix/androidsdk/ssl/PinnedTrustManager.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public void checkServerTrusted(X509Certificate[] chain, String s) throws CertificateException {
+        try {
+            if (mDefaultTrustManager != null) {
+                mDefaultTrustManager.checkServerTrusted(
+                        chain, s
+                );
+            }
+        } catch (CertificateException e) {
+            // If there is an exception we full back to checking fingerprints
```
Kegsay: `s/full/fall/`

Kegsay:
LGTM other than comments.

matrix-sdk/src/main/java/org/matrix/androidsdk/HomeserverConnectionConfig.java
erikjohnston: Since we tend to store `HomeserverConnectionConfig` I would prefer that they remained fairly immutable for now. Although the fact you can set credentials might undermine that a bit.

matrix-sdk/src/main/java/org/matrix/androidsdk/ssl/CertUtil.java
erikjohnston: Probably, although javadocs claim "SHA256" must always be supported. At the very least I should comment that.

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/login/Credentials.java
```diff
@@ -15,11 +15,41 @@
...
+        creds.accessToken = obj.getString("access_token");
+        return creds;
+    }
+
+    @Override
+    public String toString() {
+        return "Credentials{" +
+                "userId='" + userId + '\'' +
+                ", homeServer='" + homeServer + '\'' +
+                ", accessToken.length='" + accessToken.length() + '\'' +
```
Kegsay: This looks like it can NPE if `accessToken` is `null`.

Kegsay:
LGTM :)

PR #26, opened 2015-09-29
Support m.login.token

erikjohnston:
Support `m.login.token` login flow.

https://github.com/matrix-org/matrix-doc/blob/master/specification/1-client_server_api.rst#token-based


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/client/LoginRestClient.java
```diff
@@ -75,4 +78,47 @@ public void success(JsonObject jsonObject, Response response) {
             }
         });
     }
+
+    /**
+     * Attempt a user/password log in.
+     * @param user the user name
+     * @param token the password
```
Kegsay: Is "password" really the right word here?

erikjohnston: Probably not

erikjohnston: done

Kegsay:
LGTM

PR #30, opened 2016-01-15
Add remote search

ylecollen:


ylecollen:
MessagesAdapter and MatrixMessageListFragment support null mRoom case.

Such behaviour is useful to display messages whereas the room is not known or dedicated to each message (e.g a search).

PR #34, opened 2016-02-16
Update gradle wrapper, built tools, roboelectric

mitchhentges:


mitchhentges:
Agh, this should go into `develop`

PR #35, opened 2016-02-16
Update gradle wrapper, built tools, roboelectric

mitchhentges:
* Roboelectric plugin not required for `3.0`
* `@Config` not necessary. TargetSdk is used. Why test `18` when we target `21`?


ara4n:
There seems to be some concern here about the newer targets not introducing regressions on Android's side - we'll test things against 21 to see if there was a good reason for the 18 emulation, and then merge.  thanks!

mitchhentges:
I finished the bump to SDK 23, [because the build tools have to match the SDK](https://github.com/matrix-org/matrix-android-console/pull/17#issuecomment-185317571), and build tools 21 aren't available anymore.

I removed a bunch of unused imports, because some of them included classes that don't exist in SDK 23

matrix-sdk/src/main/res/values/strings.xml
```diff
@@ -27,7 +27,7 @@
     <string name="notice_avatar_changed_too">(avatar was changed too)</string>
     <string name="notice_room_name_removed">%1$s removed the room name</string>
     <string name="notice_room_topic_removed">%1$s removed the room topic</string>
-    <string name="notice_event_redacted">redacted %$1s </string>
+    <string name="notice_event_redacted">redacted %1$s </string>
```
mitchhentges: Fix lint error

ylecollen:
We updated the SDK so this issue can be closed.

PR #36, opened 2016-02-16
[SYAND-171] Viewing app after network loss can temporarily break push notifications

mitchhentges:
https://matrix.org/jira/browse/SYAND-171

### DIFF
--- a/build_alpha.sh
+++ b/build_alpha.sh
@@ -1,3 +1,0 @@
-#!/bin/bash
-./gradlew clean build
-cp app/build/outputs/apk/app-alpha-matrixorg.apk ./alpha.apk
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -1,11 +1,14 @@
 Joachim reigle <Joachim at matrix.org>
- * First implementations of the android client
+ * First implementation of the android client
 
 Kegan Dougal <kegan at matrix.org>
-  * Developpers
+ * Developer
 
 Yannick Le Collen <yannick at matrix.org>
-  * Developpers
+ * Developer
+
+Mitchell Hentges <mitchhentges at protonmail.com>
+ * Developer
 
 Matthew Hodgson <matthew at matrix.org>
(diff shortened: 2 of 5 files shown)


ara4n:
woo - thanks for that! are the build changes deliberate (removing build_alpha and abort on lint errors)? will let @ylecollen comment on the actual fix :) please feel free to add yourself to AUTHORS.rst too!

mitchhentges:
Gah, some of those build changes were deliberate. I was so excited that the small fix worked (after many hours of debugging) that I didn't check my PR after publishing

ylecollen:
I agree your fixed some sentences, removed some comments and so on

But your update does not fix the issue.

ylecollen:
This new update makes more sense but i'm afraid there is not the Kegan's issue (he talked about catchup in the logs).

I'm afraid this update does not fix all the cases.

build_alpha.sh
```diff
@@ -1,3 +0,0 @@
-#!/bin/bash
-./gradlew clean build
-cp app/build/outputs/apk/app-alpha-matrixorg.apk ./alpha.apk
```
ara4n: Why is build_alpha.sh being removed here?

mitchhentges: Because this is the SDK  project, and it can't produce an apk

ara4n: right - yup, looks like it's stale from before -console got split out into a separate project. but *please* do separate random stuff like this into seperate PRs in future, as it makes reviewing and merging much easier.

ara4n:
Right - conclusion seems to be that not resetting `isReady` every time that we try to query the push rules is a valid fix for SYAND-171: thanks @mitchhentges.  I'm slightly surprised that the code wasn't already explicitly handling the errors on the API callback, but i'm going to assume that `SimpleApiCallback` handles them sensibly.

Let's merge it. In future we'll want to improve how pushrules refresh are handled in general (especially when the server finally has a way to tell us when to update them!)

PR #75, opened 2016-09-06
Warning message when encrypted content is received (Fix #455)

pedroGitt:
Add specific message ("Encrypted message") when an encrypted message (m.room.encrypted) is received.
Should fix [#455](https://github.com/vector-im/vector-android/issues/455)


ylecollen:
LGTM

PR #77, opened 2016-09-07
Improve_notifications_management

ylecollen:
-> improve the compatibility between the web and the android client

The web client has 4 notification levels

1- All messages (lout)
2- All messages
3- Mention only
4- Mute

the android client has only 2 levels
1- Mute
2- Unmute

Web -> mobile
"All messages " and "All messages(loud)" will seen as Unmute
"Mention only" and "mute" will be seen as mute

Mobile update
Mute a room will set the "Mention only" rule
Unmute a room will set the "All messages" rule.


pedroGitt:
:+1:  LGTM

PR #80, opened 2016-09-12
fix_invalid_messages_indexes

ylecollen:
In some cases, the getView index is out ouf bound.
BY default, it used to provide the latest recycled one whereas it should warn the parent fragment that something went wrong.


pedroGitt:
LGTM

PR #81, opened 2016-09-20
Should fix

ylecollen:
https://github.com/vector-im/vector-android/issues/529
the unread notified messages are not properly cleared when the network connection is lost / unstable


pedroGitt:
LGTM

PR #82, opened 2016-09-20
Fix invalid member info preview #536

ylecollen:
should fix 
https://github.com/vector-im/vector-android/issues/536
The room preview does not always display the right member info


pedroGitt:
LGTM

PR #84, opened 2016-09-22
Fix OOM while saving

ylecollen:
Should fix issue
https://github.com/vector-im/vector-android/issues/540
All the store data is lost if there is an OOM error while saving it


pedroGitt:
LGTM

PR #85, opened 2016-09-23
fix_offline_retries_issues

ylecollen:
In some offline conditions, the unsent events were stuck when the network connection was retrieved.


pedroGitt:
:+1:  LGTM

PR #90, opened 2016-11-02
E2e

ylecollen:
First implementation of to support e2e
-> use libolm 0.1.3
-> works only with the previous riot app 
-> some unitary testes have been added to test it.


pedroGitt:
**General**
 0. How far should we go in terms of defensive programming approach to prevent leaking critical data in the stack? (see ```MXCrypto```)
 1. magic string: ```"ed25519"```, possible candidate ```OlmAccount.JSON_KEY_FINGER_PRINT_KEY```
 2. magic string: ```"curve25519"```, possible candidate ```OlmAccount.JSON_KEY_IDENTITY_KEY```
 3. magic string: ```"curve25519"```,  possible candidate ```OlmAccount.JSON_KEY_ONE_TIME_KEY``` when appropiate

**```MXCrypto```**
 1. ```startUploadKeysTimer()``` line 214:  magic number 5 (ex: ```ONE_TIME_KEY_GENERATION_MAX_NUMBER```)
 2. ```decryptEvent ()```:  where it is used, the return code is not checked (```null``` can be returned)
 3. ```decryptEvent ()``` line 1017:  ```result``` could be null at this point..
 4. ```uploadKeys ()``` line 378:   ```oneTimeKeyCountsForAlgorithm()``` returns an int, so ```keyCount``` can be used directly
 5. ```setEncryptionInRoom()```: return code is never tested (ex: in ```encryptEventContent()```)
**```MXOlmDevice```**
 6. ```signJSON ()```  - typo:  JSONDictinary => JSONDictionary

**```CryptoApi```**
 1. ```sendToDevice ()```:  ```random``` parameter may be it could be renamed in ```randomTransactionId```? (*transaction identifier is a particular access token. It only serves to identify new requests from retransmits*)

PR #92, opened 2016-11-04
replaced apache with HttpURLConnection

Denis10:
increased target sdk version


ylecollen:
I merged your UrlPostTask update but not your target updated : we have to keep the compliancy with the former console application.

PR #93, opened 2016-11-04
prevent crash on kitkat in GZIPOutputStream

Denis10:
There was fantom crash on kitkat devices

Fatal Exception: java.lang.IllegalStateException: attempt to use Deflater after calling end
       at java.util.zip.Deflater.checkOpen(Deflater.java:476)
       at java.util.zip.Deflater.deflateImpl(Deflater.java:236)
       at java.util.zip.Deflater.deflate(Deflater.java:232)
       at java.util.zip.DeflaterOutputStream.flush(DeflaterOutputStream.java:193)
       at java.io.FilterOutputStream.flush(FilterOutputStream.java:88)
       at java.io.DataOutputStream.flush(DataOutputStream.java:63)
       at java.io.ObjectOutputStream.flush(ObjectOutputStream.java:462)
       at java.io.ObjectOutputStream.close(ObjectOutputStream.java:337)
       at org.matrix.androidsdk.data.MXFileStore.saveRoomState(SourceFile:1032)

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/data/MXFileStore.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/data/MXFileStore.java
@@ -1222,7 +1222,12 @@ public class MXFileStore extends MXMemoryStore {
             if (null != room) {
                 long start1 = System.currentTimeMillis();
                 FileOutputStream fos = new FileOutputStream(roomStateFile);
-                GZIPOutputStream gz = new GZIPOutputStream(fos);
+                GZIPOutputStream gz = null;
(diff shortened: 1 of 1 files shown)


ylecollen:
Why only kitkat devices ?

bmarty:
Before KitKat, the `syncFlush` parameter does not exist.
On Kitkat, it is set to true
After Kitkat it is set to false by default

So I assume we can merge this PR, by changing also the other call to the GZIPOutputStream() constructor in MxFileCryptoStore class

bmarty:
Close because replaced by #279

PR #95, opened 2016-11-14
Fix crash when getDirectChatRoomsDict() returns null

pedroGitt:
In some accounts ```store.getDirectChatRoomsDict()``` returns null, a sanity check was missing


ylecollen:
LGTM

PR #98, opened 2016-11-16
Add attachments encryption

ylecollen:


pedroGitt:
**```FileMessage```**
1. Naming rule for field ```file``` change to something like ```mEncryptedFile```
2. ```getUrl()```: missing an ```else``` case

**```ImageMessage```**
1. Naming rule for field ```file``` change to something like ```mEncryptedFile```
2. Naming rule for field ```thumbnail_file``` change to something like ```mEncryptedThumbnailFile```
3. consider changing visibility to private for ```url, thumbnailUrl, file, thumbnail_file```?
4. ```getUrl()```: missing an ```else``` case

**```VideoMessage```**
1. same remarks a for ```ImageMessage```

**```MatrixMessageListFragment```**
1. ```uploadFileContent()``` – line 1056: magic string ("application/octet-stream") for the mime type applied to encrypted content (rather something like ```MIMETYPE_ENCRYPTED_STREAM```) 
 
**```MessagesAdapter```**
1. ```managePendingImageVideoDownload()``` line 1866:  consider calling ```getUrl()``` instead of direct access to ```url```

**```MXMediaDownloadWorkerTask```**
1. ```doInBackground()``` line 748:  is it a relevant line?

**```MXEncryptedAttachments```**
1. Magic number 32 * 1024 used twice: consider ```private final int CRYPTO_BUFFER_READ_SIZE = 32*1024;```
2. many magic strings used at least twice (```sha256```, ```oct```, ```A256CTR```, ```AES/CTR/NoPadding```…), consider ```final String``` constants?
3. ```encryptAttachment ()``` line 64:  consider renaming ```ivBytes``` to ```initVectorBytes[]``` or update comment to ```// generate a random initialization vector```
4. ```encryptAttachment ()``` line 73:  consider renaming ```cipher``` to ```encryptCipher```?
5. ```encryptAttachment ()``` line 83:  why not considering ```while(-1 != (read=attachmentStream.read(data)) ```? and remove line 87
6. ```encryptAttachment ()``` line 91:  check if this line is required, seems ```messageDigest.update()``` might be  called twice on the last read loop
7. ```decryptAttachment()```:  all same previous remarks as for ```encryptAttachment ()```
8. ```decryptAttachment()```:  consider renaming ```cipher``` to ```decryptCipher```?
9. ```decryptAttachment()``` line 199 and line 208:  code portions (```outStream.close();..```) can be refactored at the end line 215

PR #99, opened 2016-11-17
rotate_megolm_session

ylecollen:
Like the web app, rotation are based on default values defined by the spec. The code does not handle parameters that can come with the m.room.encryption event.


matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/algorithms/megolm/MXMegolmEncryption.java
```diff
@@ -616,10 +632,28 @@ private void processPendingEncryptions() {
...
+    /**
+     * Check if it's time to rotate the session
+     * @return true if a rotation is required
+     */
+    private boolean needsRotation() {
+        boolean needsRotation = false;
+        long sessionLifetime = System.currentTimeMillis() - mCreationTime;
+
+        if ((mUseCount > mSessionRotationPeriodMsgs) || (sessionLifetime > mSessionRotationPeriodMs)) {
+            Log.d(LOG_TAG, "## needsRotation() : Rotating megolm session after " + mUseCount + " messages, " + " ms");
```
manuroe: + sessionLifetime is missing

PR #110, opened 2016-11-30
Fix [Direct Message] DM marking disappear on particular circumstance …

pedroGitt:
…#701


pedroGitt:
The idea is tomerge the historical 1:1 rooms in the account_data on the first time it is detected

ylecollen:
LGTM

PR #112, opened 2016-12-01
launch_crypto_offline

ylecollen:
-> the crypto is started in a background thread : it does not anymore lock the initial sync.


ylecollen:
LGTM

PR #117, opened 2016-12-06
distinguish_users_with_no_dev

ylecollen:


matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXCrypto.java
```diff
@@ -566,14 +566,13 @@ public void downloadKeys(List<String> userIds, boolean forceDownload, final ApiC
                 for (String userId : userIds) {
 
                     Map<String, MXDeviceInfo> devices = mCryptoStore.devicesForUser(userId);
-                    boolean isEmpty = (null == devices) || (devices.size() == 0);
 
-                    if (!isEmpty) {
-                        stored.setObjects(devices, userId);
+                    if ((null == devices) || forceDownload) {
```
manuroe: forceDownload check is useless here

PR #124, opened 2016-12-20
Add video call frame update

ylecollen:
Add a method in the calls listener to be warned that the local preview size has been updated.


ylecollen:
LGTM

PR #125, opened 2017-01-11
Fix issue #799 Corrupted OlmSessions

ylecollen:
The e2e data were saved asynchronously so if the application crashed or was killed while saving them, the user could fail to read his own messages


ylecollen:
LGTM

PR #129, opened 2017-02-01
Add crypto import export

ylecollen:
Add the e2e keys import / export


matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/PBKDF2KeyImpl.java
```diff
@@ -0,0 +1,258 @@
+/*
+ * Copyright (c) 2005, 2010, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
```
manuroe: Did you check it is compatible with our license?

matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXMegolmExportEncryption.java
```diff
@@ -0,0 +1,362 @@
...
+        byte[] deriveKey = deriveKeys(salt, kdf_rounds, password);
+
+        Cipher decryptCipher = Cipher.getInstance("AES/CTR/NoPadding");
+
+        SecretKeySpec secretKeySpec = new SecretKeySpec(getAesKey(deriveKey), "AES");
+        IvParameterSpec ivParameterSpec = new IvParameterSpec(iv);
+        decryptCipher.init(Cipher.ENCRYPT_MODE, secretKeySpec, ivParameterSpec);
+
+        ByteArrayOutputStream outStream = new ByteArrayOutputStream();
+        outStream.write(decryptCipher.update(data.getBytes("UTF-8")));
```
manuroe: To save RAM, it may worth to cipher the input data by packets rather than the whole input buffer in a single pass.

PR #132, opened 2017-02-16
Add get key request support

ylecollen:
Add the e2e devices list refresh from the /sync request

Should fix https://github.com/vector-im/riot-web/issues/2305


matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXCrypto.java
```diff
@@ -1786,6 +1826,45 @@ public void run() {
...
     /**
+     * Provides the list of e2e rooms
+     * @return the list of e2e rooms
+     */
+    private List<Room> getE2eRooms() {
+        List<Room> rooms = new ArrayList<>(mSession.getDataHandler().getStore().getRooms());
+        List<Room> e2eRooms = new ArrayList<>();
+
+        for(Room r : rooms) {
+            if (r.isEncrypted()) {
```
manuroe: matrix-js-sdk and matrix-ios-sdk ignore any rooms which the user has left

PR #139, opened 2017-02-28
Add blacklisting unverified devices

ylecollen:
https://github.com/matrix-org/matrix-js-sdk/pull/336


matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXCrypto.java
```diff
@@ -2322,4 +2323,208 @@ public void onUnexpectedError(Exception e) {
...
+                });
+            }
+        });
+    }
+
+    /**
+     * Tells whether the client should ever send encrypted
+     * messages to unverified devices.
+     * This function must be called in the getEncryptingThreadHandler() thread.
+     * @return true to unilaterally blacklist all unverified devices.
```
manuroe: It may worth to indicate the default value.

PR #145, opened 2017-03-07
fix_otk_sending_order

ylecollen:
Start the crypto only there is no more device events catchup


manuroe:
The current code sends the to-device announcement event before uploading OTKs. This what js-sdk does but other devices will not be able to download these keys if they react to quickly to this annoucement.

PR #151, opened 2017-03-14
Update login api

sabrinaj:


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/client/LoginRestClient.java
```diff
@@ -124,9 +124,9 @@ public void loginWithUser(final String user, final String password, final ApiCal
         final String description = "loginWithUser : " + user;
```
ylecollen: Missing Vector Ltd

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/login/PasswordLoginParams.java
```diff
@@ -31,4 +53,41 @@
 
     // A display name to assign to the newly-created device
     public String initial_device_display_name;
+
+    public void setUserIdentifier(@NonNull final String username, @NonNull final String password) {
```
ylecollen: No function header ?

PR #154, opened 2017-03-16
Add new public rooms api

ylecollen:
Update the SDK to manage the new public rooms search APIs


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/PublicRoomsParams.java
```diff
@@ -19,15 +20,33 @@
...
     public String server;
 
-    /** Maximum number of entries to return **/
+    /**
+     * the third party instance id
+     */
+    public String thirdPartyInstanceId;
+
+    /**
+     * tell if the server must be done in all the connected networks.
```
manuroe: tell if the **query**?

PR #156, opened 2017-03-20
fix_invalid_member_avatar_url

ylecollen:
Add a RoomMember.getAvatarUrl method to check if the url is valid.

should fix https://github.com/vector-im/riot-android/issues/1042

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/RoomMember.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/RoomMember.java
@@ -18,18 +18,20 @@ package org.matrix.androidsdk.rest.model;
 
 import android.text.TextUtils;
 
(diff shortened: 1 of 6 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/RoomMember.java
```diff
@@ -166,6 +168,20 @@ public void setInviterId(String userId) {
         mInviter = userId;
     }
 
+    public String getAvatarUrl() {
+        // allow only url which starts with mxc://
+        if ((null != avatarUrl) && !avatarUrl.startsWith(ContentManager.MATRIX_CONTENT_URI_SCHEME)) {
```
Matrixcoffee: This test will fail if the avatar url starts with MXC: (or Mxc:, etc.) for some odd reason. I don't know how bad that is, but it's something to consider.

PR #170, opened 2017-05-05
Update dependencies

acq:
Updates :
- Android Gradle plugin
- compile sdk version
- target sdk version
- support libraries
- Gson
- fixes the version of Mockito

And the biggest part of the update is migrating all networking code from Retrofit 1 to Retrofit 2 APIs (which implied to rework a lot of APIs internal to the SDK).

I tested the SDK on the Riot codebase, and if you update the versions of the depencies in its build.gradle, it should work seamlessly


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/callback/RestAdapterCallback.java
```diff
@@ -130,33 +145,36 @@ public void success(T t, Response response) {
...
-            mUnsentEventsManager.onEventSendingFailed(mEventDescription, mIgnoreEventTimeLifeInOffline, error, mApiCallback, mRequestRetryCallBack);
+            mUnsentEventsManager.onEventSendingFailed(mEventDescription, mIgnoreEventTimeLifeInOffline, response, exception, mApiCallback, mRequestRetryCallBack);
         } else {
-            if (error.isNetworkError()) {
+            if (exception != null && exception instanceof IOException) {
                 try {
                     if (null != mApiCallback) {
                         try {
-                            mApiCallback.onNetworkError(error);
+                            mApiCallback.onNetworkError((Exception) exception);
```
duchuyctlk: `exception` is already an instance of `Exception` class, why do you try to cast it?

acq:
My bad.
At one time, I tried to replace the RetrofitError with a Throwable, but this would have changed too much public APIs, so I back-tracked and ended up on Exception.
But I forgot to remove the cast.
Thanks for noticing: I updated the PR.

acq:
Replaced by #174

PR #175, opened 2017-05-23
Let matrix sdk be used through Maven

vincentbrison:
This PR fix the generated pom.xml when releasing the library using maven, enable the use of Jitpack or any Maven repository.

### DIFF
--- a/matrix-sdk/build.gradle
+++ b/matrix-sdk/build.gradle
@@ -100,7 +100,7 @@ dependencies {
     compile 'com.squareup.okhttp:okhttp:2.2.0'
     compile 'io.pristine:libjingle:9690@aar'
 
-    compile(name: 'olm-sdk', ext: 'aar')
+    compile 'org.matrix:olm-sdk:olm-sdk@aar'
 
     // Robolectric
     testCompile 'com.android.support.test:runner:0.3'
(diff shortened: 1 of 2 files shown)


Matrixcoffee:
This is related to vector-im/riot-android#1284, correct?

vincentbrison:
It suits the same goal. The purpose of this PR is to let the projet be Maven compatible, thus, when using `maven` plugin for gradle, it will generate valid `pom.xml` with is not the case with the current declaration of `olm-sdk` https://github.com/matrix-org/matrix-android-sdk/pull/175/commits/ab37caa5ce40f507f4674a803d2544504e8715e8.

Then, if you want to release `matrix-android-sdk` on a Maven repository, it will be easy, and even more, we can directly as a developer start using the sdk using maven from Jitpack. Jitpack let you reference commit, github release, branches.

This feature bring a huge enhancement in dependency management, since projets using the SDK will no more include the binary and declare the dependencies of this binary, needing to know from which commit came the binary. Only add the Maven dependency.

vincentbrison:
Signed-off-by: Vincent Brison vincent.brison@fabernovel.com

matrix-sdk/build.gradle
```diff
@@ -100,7 +100,7 @@ dependencies {
     compile 'com.squareup.okhttp:okhttp:2.2.0'
     compile 'io.pristine:libjingle:9690@aar'
 
-    compile(name: 'olm-sdk', ext: 'aar')
+    compile 'org.matrix:olm-sdk:olm-sdk@aar'
```
bmarty: Unfortunately, this still uses the embedded binary file libs/olm-sdk.aar (if you remove this file it does not work anymore)
Is it the expected behavior?

bmarty: Note: this is due to the lines in build.gradle
>        flatDir {
>            dir 'libs'
>        }

PR #181, opened 2017-06-23
fix_null_read_marker_after_timeline_limited

ylecollen:
Should fix
#1364 I have seen all messages in #matrix-dev (on another device) but the room gets not unbold.
#1356 The jump to latest unread item doesn't always appear.
#1309 The notification counter isn't correct


sabrinaj:
LGTM

PR #182, opened 2017-06-28
fix_read_markers_room_creation

ylecollen:
Fix the wrong read markers management when a room is created


matrix-sdk/src/main/java/org/matrix/androidsdk/adapters/MessagesAdapter.java
```diff
@@ -851,25 +851,60 @@ public MessageRow getClosestRow(Event event) {
...
+     * @param eventId
+     * @param eventTs
+     * @return closest row
+     */
+    public MessageRow getClosestRowBeforeTs(final String eventId, final long eventTs) {
+        MessageRow messageRow = getMessageRow(eventId);
+
+        if (messageRow == null) {
+            List<MessageRow> rows = new ArrayList<>(mEventRowMap.values());
+            Collections.reverse(rows);
```
sabrinaj: - The reverse is useless since the list is made from a map valuse so the order is random
- Instead, the conditions should be updated in getClosestRow

PR #217, opened 2017-11-10
Bug Fix - MXSession: getDirectChatRoomIdsList should return the room …

giomfo:
…id of the pending invites

#213

We mark here the 1:1 invites as direct chats during the server sync.


matrix-sdk/src/main/java/org/matrix/androidsdk/MXDataHandler.java
```diff
@@ -1277,7 +1298,76 @@ private void manageResponse(final SyncResponse syncResponse, final String fromTo
...
+                                            }
+
+                                            @Override
+                                            public void onUnexpectedError(Exception e) {
+                                                Log.d(LOG_TAG, "## manageResponse() : update account data failed " + e.getMessage());
+                                            }
+                                        });
+                                    }
+                                }
+                            }
```
ylecollen: the request and other things could be moved outside the loop.

ylecollen:
LGTM

PR #218, opened 2017-11-23
fix_dm_room_creation

ylecollen:
fix DM creation with  createRoom()

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
@@ -2007,14 +2007,7 @@ public class MXSession {
             // do not wait the server request echo to update the store
             getDataHandler().getStore().setDirectChatRoomsDict(params);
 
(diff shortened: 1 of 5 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
```diff
@@ -1198,74 +1196,60 @@ public boolean createDirectMessageRoom(final String aParticipantUserId, final St
...
-                            }
-                        }
+    /**
+     * Set the created room as a direct message one.
+     *
+     * @param roomId   the room id
+     * @param userIds  the user ids list
+     * @param callback the asynchronous callback
+     */
+    private void setDirectMessageRoom(final String roomId, final List<String> userIds, final ApiCallback<String> callback) {
```
giomfo: It would be better to expect only one userId in param (instead of userIds list)

giomfo: I would rename this method with "finalizeDirectMessageRoom" because it is limited to the room creation.

PR #231, opened 2018-01-30
Save pending share keys with device

ylecollen:
The pending keys requests are saved in the crypto store until they have a proper answer.
They used to be lost when the application was killed in background.


ylecollen:
Invalid base

PR #246, opened 2018-03-16
Handle pending invitations : set the room is ready for invitations

Tyuoli:
In the pending invitations case, the room sends only very little informations. That is why we can consider that it is ready.

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/data/EventTimeline.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/data/EventTimeline.java
@@ -446,6 +446,10 @@ public class EventTimeline {
             state.setDataHandler(mDataHandler);
 
             this.mBackState = this.mState = state;
+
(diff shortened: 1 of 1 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/data/EventTimeline.java
```diff
@@ -446,6 +446,10 @@ public void handleJoinedRoomSync(RoomSync roomSync, boolean isGlobalInitialSync)
             state.setDataHandler(mDataHandler);
 
             this.mBackState = this.mState = state;
+
+            // in the pending invitation case the room sends only very little informations
+            // that is why we can consider that it is ready
+            mRoom.setReadyState(true);
```
giomfo: This is not the right place to add this change.
Here the room was just joined. This is not an invite anymore

Your change should be added in `handleInvitedRoomSync`,
after the events handling

PR #264, opened 2018-05-04
Implement Reply To feature (display only)

bmarty:
https://github.com/vector-im/riot-android/issues/2238

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/util/EventDisplay.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/util/EventDisplay.java
@@ -165,7 +165,7 @@ public class EventDisplay {
                 String msgtype = (null != jsonEventContent.get("msgtype")) ? jsonEventContent.get("msgtype").getAsString() : "";
 
                 // all m.room.message events should support the 'body' key fallback, so use it.
-                text = jsonEventContent.get("body") == null ? null : jsonEventContent.get("body").getAsString();
+                text = jsonEventContent.has("body") ? jsonEventContent.get("body").getAsString() : null;
 
                 // check for html formatting
                 if (jsonEventContent.has("formatted_body") && jsonEventContent.has("format")) {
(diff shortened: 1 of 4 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/util/EventDisplay.java
```diff
@@ -162,18 +177,50 @@ public CharSequence getTextualDisplay(Integer displayNameColor) {
...
+                                }
+                            }
+                        }
+
                         // some markers are not supported so fallback on an ascii display until to find the right way to manage them
                         // an issue has been created https://github.com/vector-im/vector-android/issues/38
-                        if (!TextUtils.isEmpty(htmlBody) && !htmlBody.contains("<ol>") && !htmlBody.contains("<li>")) {
-                            text = Html.fromHtml(jsonEventContent.getAsJsonPrimitive("formatted_body").getAsString());
+                        // BMA re-enable <ol> and <li> support (https://github.com/vector-im/riot-android/issues/2184)
+                        if (!TextUtils.isEmpty(htmlBody) /*&& !htmlBody.contains("<ol>") && !htmlBody.contains("<li>")*/) {
```
giomfo: Note: You should remove the comments from the condition. Your comment above is enough

matrix-sdk/src/main/java/org/matrix/androidsdk/util/EventDisplay.java
```diff
@@ -47,20 +49,33 @@
 public class EventDisplay {
     private static final String LOG_TAG = EventDisplay.class.getSimpleName();
 
+    private static final String MESSAGE_IN_REPLY_TO_FIRST_PART = "<blockquote data-mx-reply>";
```
ara4n: Heads up that we made a last minute change to the spec here, to simplify things a bit - sorry :(  See https://github.com/matrix-org/matrix-react-sdk/pull/1889

bmarty: Ok, is it deployed yet?

ara4n: replies are a client-side feature, and as per that PR the format change is already merged and so present on riot.im/develop and will also be in the forthcoming 0.15 release later today (albeit disabled by default in labs).

matrix-sdk/src/main/java/org/matrix/androidsdk/util/EventDisplay.java
bmarty: ok, done

matrix-sdk/src/main/java/org/matrix/androidsdk/util/EventDisplay.java
bmarty: ok, fixed

PR #268, opened 2018-05-16
Bug Fix: do not allow non-mxc content URLs.

giomfo:
we should not be making requests out to whatever http urls people send us.

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
@@ -310,7 +310,8 @@ public class MXMediasCache {
                 return mContentManager.getDownloadableUrl(url);
             }
         } else {
-            return url;
+            // do not allow non-mxc content URLs: we should not be making requests out to whatever http urls people send us
+            return null;
         }
     }
 
@@ -357,7 +358,17 @@ public class MXMediasCache {
             return null;
         }
 
-        String filename = (url.startsWith("file:")) ? url : MXMediaDownloadWorkerTask.buildFileName(downloadableUrl(url, width, height), mimeType);
+        String filename;
+        if (url.startsWith("file:")) {
+            filename = url;
+        } else {
+            String downloadableUrl = downloadableUrl(url, width, height);
+            if (null != downloadableUrl) {
+                filename = MXMediaDownloadWorkerTask.buildFileName(downloadableUrl, mimeType);
+            } else {
+                return null;
+            }
+        }
 
         try {
             // already a local file
@@ -1004,6 +1015,15 @@ public class MXMediasCache {
             downloadableUrl = downloadableUrl(url, width, height);
         } else {
             downloadableUrl = downloadableUrl(url, -1, -1);
+        }
+
+        // Check whether the url is valid
+        if (null == downloadableUrl) {
+            // Nothing to do
+            if (null != imageView) {
+                imageView.setImageBitmap(defaultBitmap);
+            }
+            return null;
         }
 
         // the thumbnail params are ignored when encrypted
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
@@ -297,7 +297,7 @@ public class MXMediasCache {
     /**
      * Resolve a Matrix media content URI (in the form of "mxc://...") into an HTTP URL.
      *
-     * @param url    the matrix url, must not be null.
+     * @param url    the matrix url.
      * @param width  the expected image width
      * @param height the expected image height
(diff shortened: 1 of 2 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
```diff
@@ -310,7 +310,8 @@ private String downloadableUrl(String url, int width, int height) {
                 return mContentManager.getDownloadableUrl(url);
             }
         } else {
-            return url;
+            // do not allow non-mxc content URLs: we should not be making requests out to whatever http urls people send us
+            return null;
```
bmarty: This is quite dangerous to return null, as before it was never returning null.
I think in some cases this should provocate NPE.
Please check all the calls to this method.
Please also use `@Nullable` annotation to the return type and update the Javadoc.

-> Maybe this is not the best way to fix this issue

bmarty: Is it still working with the sticker?

giomfo: yes it is ;)

giomfo: You're right, I missed some calls of this method which involve NPE

matrix-sdk/src/main/java/org/matrix/androidsdk/util/ContentManager.java
```diff
@@ -85,33 +87,47 @@ public static String getIdenticonURL(String userId) {
...
 
     /**
-     * Get an actual URL for accessing the full-size image of the given content URI.
+     * Check whether an url is a valid matrix content url.
+     *
+     * @param contentUrl the content URL (in the form of "mxc://...").
+     * @return true if contentUrl is valid.
+     */
+    public boolean isValidMatrixContentUrl(String contentUrl) {
+        if (null != contentUrl && contentUrl.startsWith(MATRIX_CONTENT_URI_SCHEME)) {
```
bmarty: Can be simplified to only have one `return` statement (can be done later)

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
```diff
@@ -294,23 +295,20 @@ public static void clearThumbnailsCache(Context applicationContext) {
...
      *
-     * @param url    the matrix url
+     * @param url    the matrix url, must not be null.
      * @param width  the expected image width
      * @param height the expected image height
-     * @return the URL to access the described resource.
+     * @return the URL to access the described resource, or null if the url is invalid.
      */
+    @Nullable
     private String downloadableUrl(String url, int width, int height) {
```
bmarty: `@NonNull` should be added to the `url` parameter then (can be done later)

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
```diff
@@ -321,22 +319,23 @@ private String downloadableUrl(String url, int width, int height) {
      * @param size the thumbnail size.
      * @return the File if it exits.
      */
+    @Nullable
     public File thumbnailCacheFile(String url, int size) {
```
bmarty: `url` parameter cannot be null then. Add `@NonNull` to the signature

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
```diff
@@ -294,23 +295,20 @@ public static void clearThumbnailsCache(Context applicationContext) {
     }
 
     /**
-     * Convert matrix url into http one.
+     * Resolve a Matrix media content URI (in the form of "mxc://...") into an HTTP URL.
      *
-     * @param url    the matrix url
+     * @param url    the matrix url, must not be null.
```
bmarty: Why it must not be null? The nullity checks are done.

bmarty: `url` parameter should be renamed to `mxcUrl` or `matrixUrl` for clarity (but it can be applied to all other methods...)

giomfo: you're right, I will remove this comment

giomfo: I agree with you, but we should do it later ;)

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
giomfo: I will ignore this remark, because the nullity checks are done finally

PR #269, opened 2018-05-18
Manage user widget and add API to send Event to a Room

bmarty:
Necessary to send stickers to a Room

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/data/Room.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/data/Room.java
@@ -2105,7 +2105,7 @@ public class Room {
     /**
      * Update the user enabled room url preview
      *
-     * @param status the new status
+     * @param status   the new status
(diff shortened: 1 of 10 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/data/Room.java
```diff
@@ -3028,16 +3029,33 @@ private void sendTextMessage(String text, String HTMLFormattedText, String forma
...
         roomMediaMessage.setEventCreationListener(listener);
 
         mRoomMediaMessagesSender.send(roomMediaMessage);
     }
 
+    /**
+     * Send an arbitrary event.
+     *
+     * @param event
+     * @param msgType
```
manuroe: In matrix wording, it should be `eventType`. `msgType` is a field in the event content when the event type is `m.room.message`.

bmarty: Done

PR #275, opened 2018-05-30
Add deactivate account API, the same way a logout API exists

bmarty:
To ensure the session is cleared properly once the account is deactivated server side

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
@@ -2077,6 +2077,45 @@ public class MXSession {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
```diff
@@ -2076,6 +2076,45 @@ public void onUnexpectedError(Exception e) {
         });
     }
 
+    /**
+     * Deactivate the account.
+     *
+     * @param context       the application context
+     * @param type          type of authentication
+     * @param userId        current user id
```
manuroe: A MXSession is tied to a user. The userId should not appear in arguments. `mCredentials.userId` can be used instead.

PR #280, opened 2018-06-01
Fix format of /report/ API body

richvdh:
'score' should be an integer, not a list, and synapse is going to start
enforcing this soon.


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/ReportContentParams.java
```diff
@@ -24,7 +24,7 @@
 public class ReportContentParams {
 
     // The event range from -100 “most offensive” to 0 “inoffensive”.
-    public List<Integer> score;
+    public Integer score;
```
bmarty: You can use the simple `int` type (but it's not blocking)

richvdh: oh yeah. sorry my Java is rusty!

richvdh:
fixes vector-im/riot-android/issues/2301

Issue #284, opened 2018-06-14
RoomMediaMessage#listRoomMediaMessages prefers subject over text for text/plain messages


14mRh4X0r:
This causes unexpected behaviour in apps that utilize this method, e.g. vector-im/riot-android#2140.

PR #285, opened 2018-06-14
Prefer message text instead of subject

silkeh:
This is a quick and untested patch to fix issue #284.

I have attempted to retain the behaviour of reverting to `subject\ntext` when `text` is a URL.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,7 @@ Improvements:
 Bugfix:
  - Prevent crash on KitKat
  - Prevent leaking of filenames in uploads to E2EE rooms
+ - Prefer message text instead of subject
 
 API Change:
  - New API: add device_id param to LoginRestClient.loginWithUser()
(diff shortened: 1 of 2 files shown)


bmarty:
Hi @silkeh,

Thanks for this PR. Look good to me.

Could you please sign-off it as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off

Also can you please add an entry in CHANGES.rst file?

Benoît

silkeh:
> Could you please sign-off it as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off

The commit contains a sign-off already, do you want me to add it to the PR as well?

> Also can you please add an entry in CHANGES.rst file?

Done.

bmarty:
Hi @silkeh ,
No, the message on the commit is enough, I haven't seen it, my bad.
Thanks for update
Benoit

PR #290, opened 2018-06-19
bug fix on av scanner support

giomfo:
- MXMediaCache: fix the wrong imageView tag used to download asynchronously an image
- MXMediaDownloadWorkerTask: fix the download of an encrypted media
- MXMediaDownloadWorkerTask: Do not blacklist the url on error when the media is encrypted and the av scanner is enabled because the same url is used for all encrypted media.


matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediaDownloadWorkerTask.java
```diff
@@ -768,9 +779,11 @@ protected Void doInBackground(Integer... params) {
 
                 // if some medias are not found
                 // do not try to reload them until the next application launch.
-                // TODO check what happen with the commun url for encrypted content
-                synchronized (mUnreachableUrls) {
-                    mUnreachableUrls.add(mUrl);
+                // Ignore here the case where the media is encrypted and the av scanner is enabled because the same url is used for all encrypted media.
```
bmarty: "Ignore in the case" ?
I understand the test, but it's quite confusing because the comment use "and" and the if condition use "||". Consider rewrite the comment? (Twice)

giomfo: I agree, I hesitated about this. I will try to make it clear

PR #292, opened 2018-06-21
MXCrypto: Add MXCryptoConfig class to customize/configure the e2e encryption

giomfo:
when some device id are already known for them

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
@@ -2184,14 +2184,15 @@ public class MXSession {
     /**
      * Optional set of parameters used to configure/customize the e2e encryption
      */
-    private static MXCryptoConfig mCryptoConfig = null;
+    @Nullable
+    private static MXCryptoConfig sCryptoConfig;
 
     /**
      * Define the set of parameters used to configure/customize the e2e encryption
      * This configuration must be set before instantiating the session
      */
     public static void setCryptoConfig(@Nullable MXCryptoConfig cryptoConfig) {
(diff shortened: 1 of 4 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/MXSession.java
```diff
@@ -2184,12 +2184,14 @@ public void enableCryptoWhenStarting() {
     /**
      * Optional set of parameters used to configure/customize the e2e encryption
      */
-    private MXCryptoConfig mCryptoConfig = null;
-    public void setCryptoConfig(@Nullable MXCryptoConfig cryptoConfig) {
+    private static MXCryptoConfig mCryptoConfig = null;
```
bmarty: Static variable should start with a `s` and it is not necessary to write `= null`. You can add `@Nullable` above this declaration

matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXCryptoConfig.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.matrix.androidsdk.crypto;
+
+/**
+ * Class to define the parameters used to customize or configure the end-to-end crypto.
+ */
+public class MXCryptoConfig {
+    // Tell whether the event content must be encrypted for invited members too.
+    // Presently the encryption for the invited members succeeds only if you already got some deviceId.
+    public boolean mEnableCryptoForInvitedMembers = false;
```
manuroe: `mEncryptMessagesForInvitedMembers` seems more accurate, no?
You could document that, by default, we encrypt messages only for joined members.

PR #296, opened 2018-06-27
Fix regression on image loading listener.

bmarty:
Url was used instead of DownloadId.


bmarty:
It's ok to have no new entry in CHANGES.rst for this PR

PR #299, opened 2018-06-29
Remove the automatic direct chats definition

giomfo:
when the current direct chat dictionary is empty on server sync

This mechanism (named "background compatibility heuristic") may introduce wrong direct chats, and it is not relevant anymore


giomfo:
I don't think I need to mention this in CHANGES.rst

PR #301, opened 2018-07-03
Feature/fix issue before release

bmarty:
Fallback for url preview. Ok to not update CHANGES.rst

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/URLPreview.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/URLPreview.java
@@ -17,6 +17,8 @@
 
 package org.matrix.androidsdk.rest.model;
 
+import android.text.TextUtils;
+
 import java.util.Map;
 
 /**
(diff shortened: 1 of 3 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/util/JsonUtils.java
```diff
@@ -646,4 +648,19 @@ public static String convertToUTF8(String s) {
 
         return out;
     }
+
+    /**
+     * Returns a dedicated parameter as a string
+     *
+     * @param paramName the parameter name
+     * @return the string value, or null if not define or not a String
```
manuroe: *not define**d***

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/URLPreview.java
```diff
@@ -55,35 +59,29 @@
...
     /**
      * Constructor
      *
      * @param map the constructor parameters
      */
-    public URLPreview(Map<String, Object> map) {
-        mDescription = getAsString(map, OG_DESCRIPTION);
-        mTitle = getAsString(map, OG_TITLE);
-        mType = getAsString(map, OG_TYPE);
+    public URLPreview(Map<String, Object> map, String url) {
```
manuroe: is it normal that verification tools do not complain about undocumented parameter? ;)

bmarty: It used to. I've checked in my configuration and it's enabled... Maybe an issue on the IDE.

matrix-sdk/src/main/java/org/matrix/androidsdk/util/JsonUtils.java
bmarty: ok

PR #304, opened 2018-07-05
Remove DEBUG log in release builds

af-anssi:


bmarty:
Hello @af-anssi ,

Maybe a better solution would be to introduce a log level as a static integer with constant for each levels, with a setter you will use in your application, and to test the level in each log methods.

Thanks
Benoît

matrix-sdk/src/main/java/org/matrix/androidsdk/util/Log.java
```diff
@@ -17,7 +17,7 @@
 
 import android.text.TextUtils;
 
-import org.matrix.olm.BuildConfig;
```
bmarty: I haven't see that in first review :)

PR #305, opened 2018-07-05
Protect local data

af-anssi:
This patch introduces protection of data stored locally with AES GCM on Android phones starting from KitKat.
Local data on new/fresh installation will be entirely protected; on existing/updated installations, only new written data are encrypted, existing data remain unencrypted.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ Features:
  - 
 
 Improvements:
- - 
+ - Encrypt local data (PR #305)
 
 Bugfix:
  - 
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
@@ -38,8 +38,10 @@ import org.matrix.olm.OlmSession;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
+import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -162,6 +164,8 @@ public class MXFileCryptoStore implements IMXCryptoStore {
     // tell if the store is ready
     private boolean mIsReady = false;
 
+    private Context mContext;
+
     public MXFileCryptoStore() {
     }
 
@@ -216,6 +220,8 @@ public class MXFileCryptoStore implements IMXCryptoStore {
         mTrackingStatuses = new HashMap<>();
         mOlmSessions = new HashMap<>();
         mInboundGroupSessions = new HashMap<>();
+
+        mContext = context;
     }
 
     @Override
@@ -381,14 +387,17 @@ public class MXFileCryptoStore implements IMXCryptoStore {
                 }
 
                 FileOutputStream fos = new FileOutputStream(file);
-                GZIPOutputStream gz = CompatUtil.createGzipOutputStream(fos);
+                OutputStream cos = CompatUtil.createCipherOutputStream(fos, mContext);
+                GZIPOutputStream gz = CompatUtil.createGzipOutputStream(cos);
                 ObjectOutputStream out = new ObjectOutputStream(gz);
 
                 out.writeObject(object);
+                out.flush();
                 out.close();
 
                 succeed = true;
                 Log.d(LOG_TAG, "## storeObject () : " + description + " done in " + (System.currentTimeMillis() - t0) + " ms");
+
             } catch (OutOfMemoryError oom) {
                 Log.e(LOG_TAG, "storeObject failed : " + description + " -- " + oom.getMessage(), oom);
             } catch (Exception e) {
@@ -1110,8 +1119,19 @@ public class MXFileCryptoStore implements IMXCryptoStore {
             try {
                 // the files are now zipped to reduce saving time
                 FileInputStream fis = new FileInputStream(file);
-                GZIPInputStream gz = new GZIPInputStream(fis);
+                InputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
+                GZIPInputStream gz;
+
+                if (cis != null) {
+                    gz = new GZIPInputStream(cis);
+                } else {
+                    //fallback to unencrypted stream for backward compatibility
+                    Log.i(LOG_TAG, "## loadObject() : failed to read encrypted, fallback to unencrypted read");
+                    gz = new GZIPInputStream(fis);
+                }
+
                 ObjectInputStream ois = new ObjectInputStream(gz);
+
                 object = ois.readObject();
                 ois.close();
             } catch (Exception e) {
@@ -1123,7 +1143,7 @@ public class MXFileCryptoStore implements IMXCryptoStore {
                     ObjectInputStream out = new ObjectInputStream(fis2);
 
                     object = out.readObject();
-                    fis2.close();
+                    out.close();
                 } catch (Exception subEx) {
                     // warn that some file loading fails
                     mIsCorrupted = true;
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
@@ -47,8 +47,10 @@ import org.matrix.androidsdk.util.MXOsHandler;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
+import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -2303,10 +2305,12 @@ public class MXFileStore extends MXMemoryStore {
         boolean succeed = false;
         try {
             FileOutputStream fos = new FileOutputStream(file);
-            GZIPOutputStream gz = CompatUtil.createGzipOutputStream(fos);
+            OutputStream cos = CompatUtil.createCipherOutputStream(fos, mContext);
+            GZIPOutputStream gz = CompatUtil.createGzipOutputStream(cos);
             ObjectOutputStream out = new ObjectOutputStream(gz);
 
             out.writeObject(object);
+            out.flush();
             out.close();
 
             succeed = true;
@@ -2346,7 +2350,17 @@ public class MXFileStore extends MXMemoryStore {
         Object object = null;
         try {
             FileInputStream fis = new FileInputStream(file);
-            GZIPInputStream gz = new GZIPInputStream(fis);
+            InputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
+            GZIPInputStream gz;
+
+            if (cis != null) {
+                gz = new GZIPInputStream(cis);
+            } else {
+                //fallback to unencrypted stream for backward compatibility
+                Log.i(LOG_TAG, "## readObject() : failed to read encrypted, fallback to unencrypted read");
+                gz = new GZIPInputStream(fis);
+            }
+
             ObjectInputStream ois = new ObjectInputStream(gz);
             object = ois.readObject();
             ois.close();
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
@@ -16,26 +16,268 @@
 
 package org.matrix.androidsdk.util;
 
+import android.content.Context;
+import android.content.SharedPreferences;
 import android.os.Build;
+import android.preference.PreferenceManager;
+import android.security.KeyPairGeneratorSpec;
+import android.security.keystore.KeyGenParameterSpec;
+import android.security.keystore.KeyProperties;
+import android.support.annotation.Nullable;
+import android.support.annotation.RequiresApi;
+import android.util.Base64;
 
-import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.math.BigInteger;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.PrivateKey;
+import java.security.SecureRandom;
+import java.security.UnrecoverableKeyException;
+import java.security.cert.CertificateException;
+import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.RSAKeyGenParameterSpec;
+import java.util.Calendar;
 import java.util.zip.GZIPOutputStream;
 
+import javax.crypto.Cipher;
+import javax.crypto.CipherInputStream;
+import javax.crypto.CipherOutputStream;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import javax.security.auth.x500.X500Principal;
+
 public class CompatUtil {
+    private static final String TAG = CompatUtil.class.getSimpleName();
+    private static final String ANDROID_KEY_STORE_PROVIDER = "AndroidKeyStore";
+    private static final String AES_GCM_CIPHER_TYPE = "AES/GCM/NoPadding";
+    private static final int AES_GCM_KEY_SIZE_IN_BITS = 128;
+    private static final int AES_GCM_IV_LENGTH = 12;
+    private static final String AES_LOCAL_PROTECTION_KEY_ALIAS = "aes_local_protection";
+
+    private static final String RSA_WRAP_LOCAL_PROTECTION_KEY_ALIAS = "rsa_wrap_local_protection";
+    private static final String RSA_WRAP_CIPHER_TYPE = "RSA/NONE/PKCS1Padding";
+    private static final String AES_WRAPPED_PROTECTION_KEY_SHARED_PREFERENCE = "aes_wrapped_local_protection";
+
+    private static SecretKey sLocalProtectionKey;
+    private static SecureRandom sPrng;
 
     /**
      * Create a GZIPOutputStream instance
      * Special treatment on KitKat device, force the syncFlush param to false
      * Before Kitkat, this param does not exist and after Kitkat it is set to false by default
      *
-     * @param fileOutputStream the output stream
+     * @param outputStream the output stream
      */
-    public static GZIPOutputStream createGzipOutputStream(FileOutputStream fileOutputStream) throws IOException {
+    public static GZIPOutputStream createGzipOutputStream(OutputStream outputStream) throws IOException {
         if (Build.VERSION.SDK_INT == Build.VERSION_CODES.KITKAT) {
-            return new GZIPOutputStream(fileOutputStream, false);
+            return new GZIPOutputStream(outputStream, false);
         } else {
-            return new GZIPOutputStream(fileOutputStream);
+            return new GZIPOutputStream(outputStream);
         }
+    }
+
+    /**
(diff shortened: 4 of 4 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -1120,14 +1135,26 @@ private Object loadObject(File file, String description) {
                 // if the zip deflating fails, try to use the former file saving method
```
bmarty: Please update this comment

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -1110,8 +1123,10 @@ private Object loadObject(File file, String description) {
             try {
                 // the files are now zipped to reduce saving time
                 FileInputStream fis = new FileInputStream(file);
-                GZIPInputStream gz = new GZIPInputStream(fis);
+                CipherInputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
```
bmarty: `cis`can be null, please handle this case

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -381,10 +392,12 @@ private boolean storeObject(Object object, File file, String description) {
                 }
 
                 FileOutputStream fos = new FileOutputStream(file);
-                GZIPOutputStream gz = CompatUtil.createGzipOutputStream(fos);
+                CipherOutputStream cos = CompatUtil.createCipherOutputStream(fos, mContext);
```
bmarty: `cos`can be null, please handle this case

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
```diff
@@ -16,26 +16,225 @@
...
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import javax.security.auth.x500.X500Principal;
+
 public class CompatUtil {
```
bmarty: Please add Javadoc to all new methods

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
```diff
@@ -2293,13 +2296,20 @@ private boolean writeObject(String description, File file, Object object) {
...
-            out.writeObject(object);
-            out.close();
+                out.writeObject(object);
+                out.flush();
+                out.close();
 
-            succeed = true;
+                succeed = true;
+            } else {
+                Log.e(LOG_TAG, "## writeObject() failed to open cipher output stream");
```
bmarty: You should write object unencrypted for compatibility mode

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -381,14 +392,21 @@ private boolean storeObject(Object object, File file, String description) {
...
-                succeed = true;
-                Log.d(LOG_TAG, "## storeObject () : " + description + " done in " + (System.currentTimeMillis() - t0) + " ms");
+                    out.writeObject(object);
+                    out.flush();
+                    out.close();
+
+                    succeed = true;
+                    Log.d(LOG_TAG, "## storeObject () : " + description + " done in " + (System.currentTimeMillis() - t0) + " ms");
+                } else {
+                    Log.e(LOG_TAG, "## writeObject() failed to open cipher output stream");
```
bmarty: You should write object unencrypted for compatibility mode

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
af-anssi: Are you sure this is a good idea ?

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
af-anssi: Are you sure this is a good idea ?

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
bmarty: If the SDK is run on a Android 4.1, I think the cache will not work, no?

af-anssi:
I updated the code to deal with pre-KitKat devices where encryption is not implemented/easily practicable.
Fallback to unencrypted output is thus only implemented for these devices.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -392,21 +393,24 @@ private boolean storeObject(Object object, File file, String description) {
...
-                    ObjectOutputStream out = new ObjectOutputStream(gz);
-
-                    out.writeObject(object);
-                    out.flush();
-                    out.close();
-
-                    succeed = true;
-                    Log.d(LOG_TAG, "## storeObject () : " + description + " done in " + (System.currentTimeMillis() - t0) + " ms");
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+                    CipherOutputStream cos = CompatUtil.createCipherOutputStream(fos, mContext);
```
bmarty: It's strange to test API level before calling a function from `CompatUtil`.
Is it possible that `CompatUtil.createCipherOutputStream()` return type is returning an `OutputStream` instead of a `CipherOutputStream` (just changing the type to a more generic type), and for SDK_INT < KITKAT, just return the unmodified for parameter (the value of `fos` in this case)

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -1128,15 +1132,18 @@ private Object loadObject(File file, String description) {
...
                 // the files are now zipped to reduce saving time
                 FileInputStream fis = new FileInputStream(file);
-                CipherInputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
                 GZIPInputStream gz;
 
-                // if the decryption fails, try to use the former zip defating method
-                if (cis != null) {
-                    gz = new GZIPInputStream(cis);
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+                    CipherInputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
```
bmarty: Same remark than above

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
```diff
@@ -192,6 +198,7 @@ private static SecureRandom getPrng() {
      * @param context the context holding the application shared preferences
      */
     @Nullable
+    @RequiresApi(Build.VERSION_CODES.KITKAT)
```
bmarty: remove this

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
```diff
@@ -192,6 +198,7 @@ private static SecureRandom getPrng() {
...
     public static CipherOutputStream createCipherOutputStream(OutputStream out, Context context)
             throws IOException, CertificateException, NoSuchAlgorithmException,
             UnrecoverableKeyException, InvalidKeyException, InvalidAlgorithmParameterException,
```
bmarty: if API < KITKAT, just return `out`

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
```diff
@@ -236,6 +240,7 @@ public static CipherOutputStream createCipherOutputStream(OutputStream out, Cont
      * @param context the context holding the application shared preferences
      */
     @Nullable
+    @RequiresApi(Build.VERSION_CODES.KITKAT)
     public static CipherInputStream createCipherInputStream(InputStream in, Context context)
             throws NoSuchPaddingException, NoSuchAlgorithmException, CertificateException,
             InvalidKeyException, KeyStoreException, UnrecoverableKeyException, IllegalBlockSizeException,
```
bmarty: if API < KITKAT, just return `in`

bmarty:
Hello @af-anssi ,

Thanks for the update. Sorry, I have a few remark again, it's more on the form.

Once updated, we will be able to merge this PR

Benoît

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
```diff
@@ -16,26 +16,264 @@
...
+
+        out.write(iv.length);
+        out.write(iv);
+
+        return new CipherOutputStream(out, cipher);
+    }
+
+    /**
+     * Create a CipherInputStream instance.
+     * Before Kitkat, this method will return null as local storage encryption is not implemented
```
bmarty: please update this comment

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
```diff
@@ -16,26 +16,264 @@
...
+        if (prng == null) {
+            prng = new SecureRandom();
+        }
+
+        return prng;
+    }
+
+    /**
+     * Create a CipherOutputStream instance.
+     * Before Kitkat, this method will return null as local storage encryption is not implemented for
```
bmarty: Pleas update this comment

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
```diff
@@ -2336,7 +2340,17 @@ private Object readObject(String description, File file) {
...
-            GZIPInputStream gz = new GZIPInputStream(fis);
+            InputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
+            GZIPInputStream gz;
+
+            if (cis != null) {
+                gz = new GZIPInputStream(cis);
+            } else {
+                //fallback to unencrypted stream for backward compatibility
+                Log.i(LOG_TAG, "## readObject() : failed to read encrypted, fallback to unencrypted read");
+                gz = new GZIPInputStream(fis);
```
bmarty: I think that `fis` has been read a bit by `CompatUtil.createCipherInputStream()`, so we may close `fis` and open it again.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -1110,8 +1119,19 @@ private Object loadObject(File file, String description) {
...
-                GZIPInputStream gz = new GZIPInputStream(fis);
+                InputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
+                GZIPInputStream gz;
+
+                if (cis != null) {
+                    gz = new GZIPInputStream(cis);
+                } else {
+                    //fallback to unencrypted stream for backward compatibility
+                    Log.i(LOG_TAG, "## loadObject() : failed to read encrypted, fallback to unencrypted read");
+                    gz = new GZIPInputStream(fis);
```
bmarty: I think that `fis` has been read a bit by `CompatUtil.createCipherInputStream()`, so we may close `fis` and open it again.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
af-anssi: I can use  `mark`/`reset` which is less agressive; would it be ok ?

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
af-anssi: I can use  `mark`/`reset` which is less agressive; would it be ok ?

bmarty: yes

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
bmarty: yes

bmarty:
As a general remark, I understand that the strategy is to encrypt stored file.
Please confirm that you encrypt the zipped data and that you do not zip the encrypted data (which will be far less efficient).

af-anssi:
Yes.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
af-anssi: Done. Mark / reset is done in `createCipherInputStream`.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
af-anssi: Done. Mark / reset is done in `createCipherInputStream`.

bmarty:
Ok, thanks for update.

bmarty:
(when rebasing, do not hesitate to squash all the commits)

bmarty:
After a quick exchange with the New Vector team, do you think that the performance may be impacted by this change?
Also it is possible to add a static method to enable encryption of local storage (so it will be disabled by default, to not impact other client?)

for instance (in pseudo code):
```
public static MxFileStore.enableStorageEncryption() {
   sEnableEncryption = true;
}
```
That you may call in your implementation of the SDK

Thanks for feedback

af-anssi:
I tested on several phones, and it is not perceivable.

To protect the device key generated (which is the most sensitive asset), it must be enabled by default as it is generated really early. I can add a preference switch to disable encryption later (so that it will impact only files written after it is disabled, and not already written ones so that the device key -- at least -- remains encrypted), but I am really not happy with this idea.

bmarty:
Ok, thanks for feedback. We will do some tests on our side to check performance with accounts with large room.

CHANGES.rst
```diff
@@ -5,7 +5,7 @@ Features:
  - 
 
 Improvements:
- - 
+ - MXCrypto: Encrypt the messages for invited members according to the history visibility (if the option is enabled in MXCryptoConfig).
```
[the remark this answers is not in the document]
af-anssi: You are right, sorry for the inconvenience.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
```diff
@@ -2346,7 +2350,17 @@ private Object readObject(String description, File file) {
         Object object = null;
         try {
             FileInputStream fis = new FileInputStream(file);
-            GZIPInputStream gz = new GZIPInputStream(fis);
+            InputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
```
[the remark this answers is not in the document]
af-anssi: It would not be a security measure if when something wrong happen you just drop it out. If you fail to establish a HTTPS (reps. SSH) session for instance, it is not expected to fallback to HTTP (resp. TELNET).

matrix-sdk/src/main/java/org/matrix/androidsdk/data/store/MXFileStore.java
```diff
@@ -2303,10 +2305,12 @@ private boolean writeObject(String description, File file, Object object) {
         boolean succeed = false;
         try {
             FileOutputStream fos = new FileOutputStream(file);
-            GZIPOutputStream gz = CompatUtil.createGzipOutputStream(fos);
+            OutputStream cos = CompatUtil.createCipherOutputStream(fos, mContext);
```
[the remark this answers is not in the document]
af-anssi: See comment above.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -1110,8 +1119,19 @@ private Object loadObject(File file, String description) {
             try {
                 // the files are now zipped to reduce saving time
                 FileInputStream fis = new FileInputStream(file);
-                GZIPInputStream gz = new GZIPInputStream(fis);
+                InputStream cis = CompatUtil.createCipherInputStream(fis, mContext);
```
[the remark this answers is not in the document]
af-anssi: See comment above.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/cryptostore/MXFileCryptoStore.java
```diff
@@ -381,14 +387,17 @@ private boolean storeObject(Object object, File file, String description) {
                 }
 
                 FileOutputStream fos = new FileOutputStream(file);
-                GZIPOutputStream gz = CompatUtil.createGzipOutputStream(fos);
+                OutputStream cos = CompatUtil.createCipherOutputStream(fos, mContext);
```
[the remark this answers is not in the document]
af-anssi: See comment above.

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
```diff
@@ -16,26 +16,268 @@
...
+            throws KeyStoreException, CertificateException, NoSuchAlgorithmException, IOException,
+            NoSuchProviderException, InvalidAlgorithmParameterException, NoSuchPaddingException,
+            InvalidKeyException, IllegalBlockSizeException, UnrecoverableKeyException {
+        if (local_protection_key == null) {
+            final KeyStore keyStore = KeyStore.getInstance(ANDROID_KEY_STORE_PROVIDER);
+            keyStore.load(null);
+
+            Log.i(TAG, "Loading local protection key");
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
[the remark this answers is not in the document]
af-anssi: Yes, some improvements should be done to improve device upgrade as the cache is not decryptable.
A (very) complex solution could be detect the upgrade and to reencrypt all cache entries.
A simpler solution could be to remove any corrupted file, forcing it do be redownloaded and re-encrypted accordingly.

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
```diff
@@ -16,26 +16,268 @@
...
+            NoSuchProviderException, InvalidAlgorithmParameterException, IOException {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT) {
+            return in;
+        }
+
+        in.mark(4 + AES_GCM_IV_LENGTH);
+        final int iv_len = in.read();
+        if (iv_len != AES_GCM_IV_LENGTH) {
+            Log.e(TAG, "Invalid IV length " + iv_len);
+            in.reset();
```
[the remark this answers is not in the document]
af-anssi: On which version of Android do you get this error ?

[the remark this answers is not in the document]
bmarty: Android 7.0

matrix-sdk/src/main/java/org/matrix/androidsdk/util/CompatUtil.java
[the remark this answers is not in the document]
bmarty: I think we have to handle this then. In the future device may be upgraded to Android M

bmarty:
@af-anssi:
I will merge your PR in another branch and add some commits you will be able to review

giomfo:
@af-anssi: FYI @bmarty will merge your PR in an intermediate branch to apply the following changes before merging it into `develop`:
- make optional the local data encryption (this option will be enabled at the application level in tchap-android).
- handle correctly the potential upgrade from <M to M, by detecting this case and clearing the cache.
- fix the `reset` use on `InputStream` (Any suggestion on this last point is welcome).

bmarty:
Point 1: Ok
Point 2 has been implemented. Any encrypted storage on OS <M will continue to use the generated key, even it the system is now >= M
Point 3: we are now closing the stream if reset is not supported.

PR #309, opened 2018-07-06
MXCrypto: Encrypt the messages for invited members according to the h…

giomfo:
…istory visibility (if the option is enabled in MXCryptoConfig).

Related to PR #292

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXCrypto.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXCrypto.java
@@ -825,21 +825,6 @@ public class MXCrypto {
             return;
         }
 
-        final List<String> userRoomIds = new ArrayList<>();
-
-        Collection<Room> rooms = mSession.getDataHandler().getStore().getRooms();
-
-        for (Room room : rooms) {
-            if (room.isEncrypted()) {
-                RoomMember roomMember = room.getMember(userId);
-
-                // test if the user joins the room
-                if ((null != roomMember) && TextUtils.equals(roomMember.membership, RoomMember.MEMBERSHIP_JOIN)) {
-                    userRoomIds.add(room.getRoomId());
-                }
-            }
-        }
-
         getEncryptingThreadHandler().post(new Runnable() {
             @Override
             public void run() {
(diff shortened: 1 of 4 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXCrypto.java
```diff
@@ -825,21 +825,6 @@ public void setDeviceVerification(final int verificationStatus, final String dev
...
-        for (Room room : rooms) {
-            if (room.isEncrypted()) {
-                RoomMember roomMember = room.getMember(userId);
-
-                // test if the user joins the room
-                if ((null != roomMember) && TextUtils.equals(roomMember.membership, RoomMember.MEMBERSHIP_JOIN)) {
-                    userRoomIds.add(room.getRoomId());
-                }
-            }
-        }
```
bmarty: Ok to remove this dead code. Missing cleanup from 444644570ac6b424b700d38879a839247ba80f4f

matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/MXCrypto.java
```diff
@@ -2427,22 +2428,6 @@ public void onSuccess(MXUsersDevicesMap<MXDeviceInfo> devicesMap) {
...
-
```
bmarty: Ok to remove this dead code. Missing cleanup from 444644570ac6b424b700d38879a839247ba80f4f

matrix-sdk/src/main/java/org/matrix/androidsdk/data/Room.java
```diff
@@ -183,6 +183,16 @@ public IMXStore getStore() {
         return mStore;
     }
 
+    /**
+     * Determine whether we should encrypt messages for invited users in this room.
+     *
+     * @return true if we should encrypt messages for invited users.
+     */
+    public boolean shouldEncryptForInvitedMembers() {
```
bmarty: From the Room class point of view, this method is not related to encryption.
I would rename it to `isHistoryVisibleForNotJoinedMembers` or something similar
Please also update the Javadoc accordingly

matrix-sdk/src/main/java/org/matrix/androidsdk/data/Room.java
```diff
@@ -183,6 +183,16 @@ public IMXStore getStore() {
...
+        String historyVisibility = getLiveState().history_visibility;
```
bmarty: Sometimes we use `getState()` and sometimes we use `getLiveState()`. Actually `getLiveState()` is just an alias for `getState()`. We should consider keeping only one method for clarity.

giomfo: ok

matrix-sdk/src/main/java/org/matrix/androidsdk/data/Room.java
giomfo: ok

PR #315, opened 2018-07-11
Rename some classes from Bing* to Push* to match server denomination

bmarty:


bmarty:
(no impact on Riot)

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/client/PushRulesRestClient.java
```diff
@@ -17,32 +18,28 @@
...
 
     /**
-     * Retrieve the bing rules list.
+     * Retrieve the push rules list.
      *
      * @param callback the asynchronous callback.
      */
-    public void getAllBingRules(final ApiCallback<BingRulesResponse> callback) {
-        mApi.getAllBingRules().enqueue(new DefaultRetrofit2CallbackWrapper<>(callback));
+    public void getAllRules(final ApiCallback<PushRulesResponse> callback) {
```
giomfo: I would rename it `getAllPushRules`

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/PushRulesApi.java
```diff
@@ -26,18 +27,20 @@
...
-public interface BingRulesApi {
+public interface PushRulesApi {
 
-    // TODO Add Javadoc
+    /**
+     * Get all push rules
+     */
     @GET("pushrules/")
-    Call<BingRulesResponse> getAllBingRules();
+    Call<PushRulesResponse> getAllRules();
```
giomfo: I would rename it getAllPushRules()

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/client/PushRulesRestClient.java
bmarty: Other methods are like this (ex: `deleteRule()`) so I think it's more coherent.

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/PushRulesApi.java
bmarty: Other methods are like this (ex: `deleteRule()`) so I think it's more coherent.

PR #336, opened 2018-08-09
remove PieFractionView from android sdk

dkanada:
I wanted to change the PieFractionView colors in the android client but it seems the logic is handled in this library. Let me know if I broke something, I actually couldn't get the SDK to compile in Android Studio because of a 'filename too long' error during compilation that I didn't want to debug.


bmarty:
Hi @dkanada ,

thanks for the PR.

Please rebase on /develop to fix the compilation issue.

The class `PieFractionView` should be in Riot and not the in the Matrix SDK. Feel free to move it to Riot in an appropriate package (so this PR can remove this class and open another PR in Riot with this class modified as you want)

Benoît

bmarty:
Ok, LGTM, please add this line in `CHANGES.rst` file in the section `API Change`:

"Remove PieFractionView class from the Matrix SDK. This class is now in Riot sources"

in case other SDK clients uses it.

Also please Signoff, else I cannot merge this PR

Thanks

bmarty:
Also, I think the resources file `colors.xml` can also be removed. Please do so

dkanada:

PR #337, opened 2018-08-11
Allow '@' in room alias

splitowo:
Allows the '@' character in the room alias. Unfortunately, the Matrix spec doesn't define allowed characters or grammar for room aliases, but '@' sure seems to be valid.

Fixes: https://github.com/vector-im/riot-android/issues/2079
Matrix spec: https://matrix.org/docs/spec/appendices.html#common-identifier-format

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Improvements:
  -
 
 Bugfix:
- -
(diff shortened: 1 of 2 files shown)


splitowo:
Hi, thanks for having a look. The changelog entry is added to the branch.

Signed-off-by: Zbigniew Brzeziński <brzezinski.z@gmx.com>

PR #338, opened 2018-08-15
Do not nullify room name or topic in case it is empty

splitowo:
Hi,

I was looking at https://github.com/vector-im/riot-android/issues/2345
The name is indeed missing from the API request. This is caused by Gson, which, by default, ignores null objects and does not serialize them. The name String is null in this case. Gson null object serialization can be enabled here, using the last parameter in the superclass constructor:
https://github.com/matrix-org/matrix-android-sdk/blob/develop/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/client/RoomsRestClient.java#L71
This, however, doesn't fix the issue as the server complains the name is not of the String type. The actual JSON sent is `{"name":null}`.

I think the correct solution would be to skip nullifying the name String. The JSON is then `{"name":""}` and the room is assigned the original, default name. I was looking at other usages of the name String so this change doesn't break anything, but all usages seem to be checked with `TextUtils.isEmpty()` anyway, so it should be safe.

There is the same story with room topic.

Fixes https://github.com/vector-im/riot-android/issues/2345
Signed-off-by: Zbigniew Brzeziński brzezinski.z@gmx.com

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@ Improvements:
 
 Bugfix:
  - Room aliases including the '@' character are now recognized as valid (vector-im/riot-android#2079)
+ - Room name and topic can be now set back to empty
 
 API Change:
  - Remove PieFractionView class from the Matrix SDK. This class is now in Riot sources (#336)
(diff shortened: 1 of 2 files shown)


bmarty:
Hi @splitowo,

Perfect! Thanks

Benoît

PR #340, opened 2018-08-22
Feature/issues

bmarty:
- MXMediasCache.createTmpMediaFile() methods are renamed to createTmpDecryptedMediaFile()
 - MXMediasCache.clearTmpCache() method is renamed to clearTmpDecryptedMediaCache()
 - Add MXMediasCache.moveToShareFolder() to move a tmp decrypted file to another folder to prevent deletion during sharing. New API MXMediasCache.clearShareDecryptedMediaCache() can be called when the application is resumed. (vector-im/riot-android#2530)

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
@@ -487,6 +487,9 @@ public class MXMediasCache {
                             while ((len = fis.read(buf)) != -1) {
                                 fos.write(buf, 0, len);
                             }
+
+                            fis.close();
+                            fos.close();
                         } catch (Exception e) {
                             Log.e(LOG_TAG, "## createTmpMediaFile() failed " + e.getMessage(), e);
                         }
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
@@ -347,7 +347,7 @@ public class MXMediasCache {
     }
 
     /**
-     * Return the cache file name for a media defined by its URL and its mimetype.
+     * Return the cache file name for a media defined by its URL and its mime type.
      *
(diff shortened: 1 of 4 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
```diff
@@ -517,6 +534,47 @@ public void clearTmpCache() {
         }
     }
 
+    /**
+     * Move a decrypted media file to the /share folder, to avoid this file to be deleted if in the /tmp folder
+     *
+     * @param fileToMove The file to move
```
giomfo: note: `filename` parameter is not commented here

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
```diff
@@ -517,6 +534,47 @@ public void clearTmpCache() {
...
+     * Clear the temporary shared decrypted media cache folder
+     */
+    public void clearShareDecryptedMediaCache() {
+        Log.d(LOG_TAG, "clearShareDecryptedMediaCache()");
+
+        if (mShareFolderFile.exists()) {
+            ContentUtils.deleteDirectory(mShareFolderFile);
+        }
+
+        if (!mShareFolderFile.exists()) {
```
giomfo: Why do you test if the folder exist, you just deleted it ?

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
bmarty: done

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
bmarty: I've copied what is done in `clearTmpDecryptedMediaCache()`...
I suppose `ContentUtils.deleteDirectory()` may fail for some reason.

PR #345, opened 2018-08-24
Add ability to use filter from server

krombel:
This adds Interfaces to create a Filter which can be send to the server.
This especially handles the DataSaveMode.
The Filter is therefore uploaded in lazy-load manner and used once the
server responds with a filter_id.

This also deprecates "setUseDataSaveMode" and now setSyncFilter should
be used. This is because the app might store the filter-id so the call
for the filter-id could be avoided if the app handles that directly

Signed-Off-by: Matthias Kesler <krombel@krombel.de>

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/FilterApi.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/FilterApi.java
@@ -35,7 +35,7 @@ public interface FilterApi {
      * @param body   the Json representation of a FilterBody object
      */
     @POST("user/{userId}/filter")
-    Call<FilterResponse> uploadFilter(@Path("userId") String userId, @Body Object body);
+    Call<FilterResponse> uploadFilter(@Path("userId") String userId, @Body FilterBody body);
 
     /**
      * Gets a filter with a given filterId from the homeserver
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/FilterApi.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/FilterApi.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2015 OpenMarket Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.matrix.androidsdk.rest.api;
+
+import org.matrix.androidsdk.rest.model.filter.FilterBody;
+import org.matrix.androidsdk.rest.model.filter.FilterResponse;
+
+import java.util.Map;
+
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.GET;
+import retrofit2.http.POST;
+import retrofit2.http.Path;
+
+public interface FilterApi {
+
+    /**
+     * Set some account_data for the client.
+     *
+     * @param userId   the user id
+     * @param body   the Json representation of a FilterBody object
+     */
+    @POST("user/{userId}/filter")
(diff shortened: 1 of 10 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/FilterApi.java
```diff
@@ -0,0 +1,49 @@
...
+public interface FilterApi {
+
+    /**
+     * Set some account_data for the client.
+     *
+     * @param userId   the user id
+     * @param body   the Json representation of a FilterBody object
+     */
+    @POST("user/{userId}/filter")
+    Call<FilterResponse> uploadFilter(@Path("userId") String userId, @Body Object body);
```
bmarty: `Object boby` should be of type `FilterBody`

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/FilterApi.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2015 OpenMarket Ltd
```
bmarty: Replace by Copyright 2108 (New Vector Ltd) please

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/FilterApi.java
```diff
@@ -0,0 +1,49 @@
...
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.GET;
+import retrofit2.http.POST;
+import retrofit2.http.Path;
+
+public interface FilterApi {
+
+    /**
+     * Set some account_data for the client.
```
bmarty: Please update this comment

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/client/FilterRestClient.java
```diff
@@ -0,0 +1,61 @@
+package org.matrix.androidsdk.rest.client;
```
bmarty: please add copyright block

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/filter/RoomEventFilter.java
```diff
@@ -0,0 +1,48 @@
...
+    public List<String> senders;
+
+    public List<String> types;
+
+    public List<String> rooms;
+
+    @SerializedName("not_rooms")
+    public List<String> notRooms;
+
+    public Boolean contains_url;
```
bmarty: please use @SerializeName to avoid `_` in member name

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/filter/RoomFilter.java
```diff
@@ -0,0 +1,45 @@
...
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * Represents "RoomFilter" as mentioned in the SPEC
+ * https://matrix.org/docs/spec/client_server/r0.3.0.html#post-matrix-client-r0-user-userid-filter
+ */
+public class RoomFilter {
+
+    public List<String> not_rooms;
```
bmarty: please use @SerializeName to avoid `_` in member name

krombel:
> About the model, even if well documented, we have to considered that it's not a close model. For instance, for the lazy loading, the Json filter would be
`{"room":{"state":{"lazy_load_members":true}}}`, so the `state` should be of type 
`Map<String, Object>`. So I'm afraid `FilterBody` should be only a `Map<String, Object>`. A Helper may be created to handle what's in the Matrix specification for filter.

As mentioned in #riot-android I would prefer extending the model instead of implementing `Map<String, Object>`. Nevertheless this is an matrix-sdk so it should not support arbitrary data I would say.

You can take this snippet to the branch which implements lazy loading - so people can see there as well, that the model changed a bit
```diff
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/filter/RoomFilter.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/filter/RoomFilter.java
@@ -42,4 +42,7 @@ public class RoomFilter {

     @SerializedName("account_data")
     public RoomEventFilter accountData;
+
+    @SerializedName("lazy_load_members")
+    public Boolean lazyLoadMembers;
 }
```

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/FilterApi.java
krombel: Removed the wrong line while Copy/Pasting...

PR #348, opened 2018-08-29
fix excessive whitespace on quoted messages

dkanada:
This is not an ideal solution since it will also remove newlines even if a user intended to end a message with whitespace. I just wanted to see what solution you think is best. This code is a lot more simple than whatever would have to be done to selectively remove the newlines following quoted lines.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Improvements:
  - 
 
 Bugfix:
- - 
+ - Fix excessive whitespace on quoted messages (#348)
 
 API Change:
  - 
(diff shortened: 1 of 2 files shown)


dkanada:
Here is an image of what the whitespace looks like.

quote

dkanada:
I have no idea how to integrate this into the android client manually but if you explain how I will test it out on my phone for a picture.

bmarty:
Thanks for the PR.

To test with the Android client, just open Riot in Android studio and run the script `compile_with_sdk_project.sh`

dkanada:
That was easier than expected, here is a photo after the fix.

dkanada:

bmarty:
Hello @dkanada ,

I think it's ok, because Riot removes any trailing space while sending messages, (if you like it or not by the way)

Benoît

PR #354, opened 2018-09-15
scroll to bottom no longer keeps inertia after position change

dkanada:
You can test the change by scrolling up really fast in a room, then jumping to the bottom of the adapter. The adapter will jump to the latest message but keep scrolling up, this pull request fixes that.

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/fragments/MatrixMessageListFragment.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/fragments/MatrixMessageListFragment.java
@@ -735,6 +735,8 @@ public class MatrixMessageListFragment extends Fragment implements MatrixMessage
             @Override
             public void run() {
                 mMessageListView.setSelection(mAdapter.getCount() - 1);
+                // stop any scroll inertia after the jump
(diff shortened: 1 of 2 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/fragments/MatrixMessageListFragment.java
```diff
@@ -737,6 +737,7 @@ public void run() {
                 mMessageListView.setSelection(mAdapter.getCount() - 1);
             }
         }, Math.max(delayMs, 0));
+        mMessageListView.smoothScrollBy(0, 0);
```
bmarty: Ok, I've checked and it works, thanks. Maybe move this new line above `mMessageListView.postDelayed(new Runnable() {` and add a comment, for instance: `// Stop the current scrolling of the list if any`

dkanada: I actually moved it inside the postDelayed method. If there was a long delay and it was outside the method the adapter would stop, wait a bit, then skip to the bottom. That might be confusing. Should be fixed now though.

PR #356, opened 2018-09-18
Feature/hhs notice soft

ganfra:
Allows to handle the first point of https://github.com/vector-im/riot-android/issues/2547

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/data/DataRetriever.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/data/DataRetriever.java
@@ -1,13 +1,13 @@
-/* 
+/*
  * Copyright 2014 OpenMarket Ltd
  * Copyright 2018 New Vector Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -87,6 +87,24 @@ public class DataRetriever {
     }
 
     /**
+     * Get the event associated with the eventId and roomId
+     * Look in the store before hitting the rest client.
+     *
+     * @param store    the store to look in
+     * @param roomId   the room Id
+     * @param eventId  the eventId
+     * @param callback the callback
+     */
+    public void getEvent(final IMXStore store, final String roomId, final String eventId, final ApiCallback<Event> callback) {
+        final Event event = store.getEvent(eventId, roomId);
+        if (event == null) {
+            mRestClient.getEvent(roomId, eventId, callback);
+        } else {
+            callback.onSuccess(event);
+        }
+    }
+
+    /**
      * Trigger a back pagination for a dedicated room from Token.
      *
      * @param store    the store to use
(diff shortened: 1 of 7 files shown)


bmarty:
Can you fix the Travis error please?

matrix-sdk/src/main/java/org/matrix/androidsdk/data/DataRetriever.java
```diff
@@ -246,20 +264,21 @@ public void onUnexpectedError(Exception e) {
      * @param token    the start token.
      * @param callback the callback
      */
-    private void forwardPaginate(final IMXStore store, final String roomId, final String token, final ApiCallback<TokensChunkResponse<Event>> callback) {
+    private void forwardPaginate(final IMXStore store, final String roomId,
```
bmarty: Either put all parameters in one line, or put every parameter on its own line (this rule should be applied for functions with 3+ parameters - to be discussed). I do not like to split them like this, I think its harder to read.
Having one parameter per line maybe another solution, especially if we use `final` and nullability annotations.
And Github comparaison/review may be easier because this limited tool only compares and allows comment on full lines.

matrix-sdk/src/main/java/org/matrix/androidsdk/data/DataRetriever.java
```diff
@@ -159,82 +177,82 @@ public void run() {
...
+                        private void logErrorMessage(String expectedToken, String errorMessage) {
+                            Log.e(LOG_TAG, "## backPaginate() failed : roomId " + roomId + " token " + token + " expected " + expectedToken + " with " + errorMessage);
                         }
 
-                        Log.d(LOG_TAG, "## backPaginate() succeed : roomId " + roomId + " token " + token + " got " + events.chunk.size());
-                        callback.onSuccess(events);
-                    }
-                }
+                        @Override
+                        public void onNetworkError(Exception e) {
```
bmarty: GitHub comparison tools is so limited... It's seems that there is only new spaces and this tool consider the whole code has changed. Or is there something I miss?

matrix-sdk/src/main/java/org/matrix/androidsdk/data/DataRetriever.java
```diff
@@ -272,7 +291,8 @@ public void onSuccess(TokensChunkResponse<Event> events) {
      * @param callback  the onComplete callback
      */
     public void paginate(final IMXStore store, final String roomId, final String token,
-                         final EventTimeline.Direction direction, final ApiCallback<TokensChunkResponse<Event>> callback) {
+                         final EventTimeline.Direction direction,
```
bmarty: same remark on parameters formatting

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/RoomPinnedEventsContent.java
```diff
@@ -0,0 +1,39 @@
...
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Content of a m.room.pinned_events type event
+ */
+public class RoomPinnedEventsContent implements Serializable {
+
+    public List<String> pinned;
```
bmarty: Please add a comment like below:
`// List of eventIds of pinned events` or something similar

matrix-sdk/src/main/java/org/matrix/androidsdk/util/JsonUtils.java
```diff
@@ -478,6 +479,19 @@ public static RoomCreateContent toRoomCreateContent(final JsonElement jsonElemen
...
+     *
+     * @param jsonElement the json to convert
+     * @return a RoomPinnedEventsContent
+     */
+    public static RoomPinnedEventsContent toRoomPinnedEventsContent(final JsonElement jsonElement) {
+        return toClass(jsonElement, RoomPinnedEventsContent.class);
+    }
+
+
+
```
bmarty: too many new lines :)

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/ServerNoticeUsageLimitContent.java
```diff
@@ -0,0 +1,32 @@
...
+package org.matrix.androidsdk.rest.model;
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * Content of a m.server_notice.usage_limit_reached type event
+ */
+public class ServerNoticeUsageLimitContent {
+
+    public String limit;
```
bmarty: What sort of data will contain `limit`? Can you please comment it?

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/Event.java
```diff
@@ -93,6 +93,7 @@
     public static final String EVENT_TYPE_ROOM_KEY_REQUEST = "m.room_key_request";
     public static final String EVENT_TYPE_FORWARDED_ROOM_KEY = "m.forwarded_room_key";
     public static final String EVENT_TYPE_URL_PREVIEW = "org.matrix.room.preview_urls";
+    public static final String EVENT_TYPE_SERVER_NOTICE_USAGE_LIMIT = "m.server_notice.usage_limit_reached";
```
bmarty: Is it the best place to add this String? Maybe move it to ServerNoticeUsageLimitContent class?

matrix-sdk/src/main/java/org/matrix/androidsdk/data/DataRetriever.java
ganfra: Didn't touch this code, automatic reformat I guess

matrix-sdk/src/main/java/org/matrix/androidsdk/data/DataRetriever.java
ganfra: Yes sorry for that, automatic reformat. 
I think we should keep in one line if it fits the limit of 160char.
Otherwise, like you said, one per line.

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/Event.java
ganfra: Yes sorry, it's a miss from me, at start it was supposed to be a true event type.

PR #374, opened 2018-10-10
Improve code

bmarty:


matrix-sdk/src/main/java/org/matrix/androidsdk/call/IMXCall.java
```diff
@@ -170,9 +172,9 @@
     /**
      * The call is hung up.
      *
-     * @param reason the reason
+     * @param reason the reason, or null for no reason. Reasons are used to indicate errors in the current VoIP implementation.
      */
-    void hangup(String reason);
+    void hangup(@Nullable String reason);
```
ganfra: Maybe having an enum would be better?

bmarty: Yes, maybe it as to be discussed during the spec workshop.

PR #382, opened 2018-10-17
Fix issue with room avatar for invitation

bmarty:
Also add unitary test for room avatar computation


matrix-sdk/src/main/java/org/matrix/androidsdk/data/room/RoomAvatarResolver.java
```diff
@@ -0,0 +1,71 @@
...
+    public RoomAvatarResolver(Room room) {
+        mRoom = room;
+    }
+
+    /**
+     * Compute the room avatar url
+     *
+     * @return the room avatar url
+     */
+    public String resolve() {
```
giomfo: add nullable

bmarty: done

PR #395, opened 2018-11-15
generate thumbnails for gifs rather than throw an error

dkanada:
This fixes an error that broke thumbnails for gifs, I tested it against a bunch of files and nothing else has changed as far as I know. It was quite annoying in encrypted chats when the entire file had to download for previews to work.


dkanada:

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
```diff
@@ -599,12 +599,6 @@ public String saveBitmap(Bitmap bitmap, String defaultFileName) {
             File file = new File(getFolderFile(null), filename);
             FileOutputStream fos = new FileOutputStream(file.getPath());
 
-            // We got an java.lang.IllegalStateException: Can't compress a recycled bitmap
-            if (bitmap.isRecycled()) {
-                Log.w(LOG_TAG, "Trying to compress a recycled Bitmap. Create a copy first.");
-                bitmap = Bitmap.createBitmap(bitmap);
-            }
```
bmarty: Are you sure it is necessary to remove this block? I've added it because on some device it fixes an issue (not observed on all devices).

matrix-sdk/src/main/java/org/matrix/androidsdk/util/ResourceUtils.java
```diff
@@ -171,12 +171,6 @@ public static Bitmap createThumbnailBitmap(Context context, Uri mediaUri, int ma
                         Log.e(LOG_TAG, "createThumbnailBitmap " + ex.getMessage(), ex);
                     }
                 }
-
-                // reduce the memory consumption
-                if (null != fullSizeBitmap) {
-                    fullSizeBitmap.recycle();
```
bmarty: Was it the source of the Exception?

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
dkanada: As far as I know the root cause of that issue was the other block I had removed below which I have proposed a fix for as well.

matrix-sdk/src/main/java/org/matrix/androidsdk/util/ResourceUtils.java
```diff
@@ -150,7 +150,7 @@ public static Bitmap createThumbnailBitmap(Context context, Uri mediaUri, int ma
 
                 // the bitmap is smaller that max sizes
                 if ((fullSizeBitmap.getHeight() < maxThumbHeight) && (fullSizeBitmap.getWidth() < maxThumbWidth)) {
-                    thumbnailBitmap = fullSizeBitmap;
+                    thumbnailBitmap = fullSizeBitmap.copy(Bitmap.Config.ARGB_8888, true);
```
dkanada: So right here I think the bitmap was getting passed by reference. When fullSizeBitmap is recycled below thumbnailBitmap also breaks, which causes the error you fixed in the MediaCache. This line fixes the root cause, but since the system runs its own gc we might as well throw away the manual recycle below as well as the other line that copies the bitmap in MediaCache.

dkanada:
@bmarty any thoughts?

matrix-sdk/src/main/java/org/matrix/androidsdk/db/MXMediasCache.java
```diff
@@ -599,7 +599,7 @@ public String saveBitmap(Bitmap bitmap, String defaultFileName) {
             File file = new File(getFolderFile(null), filename);
             FileOutputStream fos = new FileOutputStream(file.getPath());
 
-            bitmap.compress(Bitmap.CompressFormat.JPEG, 100, fos);
+            bitmap.compress(Bitmap.CompressFormat.PNG, 100, fos);
```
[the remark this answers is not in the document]
bmarty: I think there is an issue because we use mime type "image/jpeg" in `RoomMediaMessagesSender`.
Also filename is `.jpg` in this method.
Maybe we can considere keeping JPG for this PR and try to do better on another PR: keep PNG or JPG depending on the original image format.

dkanada:
Should be good to go now.

bmarty:
Ok, thanks for the update. I will manage to rename MXMediasCache to MXMediaCache in Riot

PR #397, opened 2018-11-19
Handle any AccountData element

bmarty:
Any Account data element, even if the type is not known is persisted. So they can be accessed at any time using the new API getAccountDataElement(). If AccountData is updated (from any device logged on the same account), IMXEventListener.onAccountDataUpdated() will be called.

This PR also removes a useless check on network connectivity. The OkHttp lib handle this properly.

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/MXDataHandler.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/MXDataHandler.java
@@ -1025,7 +1025,7 @@ public class MXDataHandler {
      */
     private void managePushRulesUpdate(List<AccountDataElement> accountDataElements) {
         for (AccountDataElement accountDataElement : accountDataElements) {
-            if (TextUtils.equals(accountDataElement.type, AccountDataRestClient.ACCOUNT_DATA_TYPE_PUSH_RULES)) {
+            if (TextUtils.equals(accountDataElement.type, AccountDataElement.ACCOUNT_DATA_TYPE_PUSH_RULES)) {
                 Gson gson = JsonUtils.getGson(false);
 
                 // convert the data to PushRulesResponse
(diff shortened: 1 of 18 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/MXDataHandler.java
```diff
@@ -1025,24 +1021,20 @@ private void manageAccountData(Map<String, Object> accountData, boolean isInitia
...
      */
-    private void managePushRulesUpdate(List<Map<String, Object>> events) {
-        for (Map<String, Object> event : events) {
-            String type = (String) event.get("type");
-
-            if (TextUtils.equals(type, "m.push_rules")) {
-                if (event.containsKey("content")) {
-                    Gson gson = JsonUtils.getGson(false);
+    private void managePushRulesUpdate(List<AccountDataElement> accountDataElements) {
+        for (AccountDataElement accountDataElement : accountDataElements) {
```
ganfra: I think type check should be done on the top method `manageAccountData` as we repeat for loops at the moment.

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/client/AccountDataRestClient.java
```diff
@@ -27,12 +27,13 @@
...
 public class AccountDataRestClient extends RestClient<AccountDataApi> {
     /**
-     * Account data types
+     * Account data known types
      */
     public static final String ACCOUNT_DATA_TYPE_IGNORED_USER_LIST = "m.ignored_user_list";
     public static final String ACCOUNT_DATA_TYPE_DIRECT_MESSAGES = "m.direct";
     public static final String ACCOUNT_DATA_TYPE_PREVIEW_URLS = "org.matrix.preview_urls";
     public static final String ACCOUNT_DATA_TYPE_WIDGETS = "m.widgets";
+    public static final String ACCOUNT_DATA_TYPE_PUSH_RULES = "m.push_rules";
```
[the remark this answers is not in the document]
bmarty: done

matrix-sdk/src/main/java/org/matrix/androidsdk/MXDataHandler.java
bmarty: done

PR #416, opened 2019-01-10
do not attempt to serialize unserializable fields on MatrixError in E…

knaeckeKami:
…vent::writeExternal

### Pull Request Checklist

* ✅ Pull request is based on the develop branch
* ✅ Pull request updates [CHANGES.rst](https://github.com/matrix-org/matrix-android-sdk/blob/develop/CHANGES.rst)
* ✅ Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

Currently, serializing Events that have a MatrixError set fails with a 
`java.io.NotSerializableException: okhttp3.ResponseBody$1` Exception.

It seems that this means, that the sending won't be retried later. 
This fixes the exception, by stripping the event of all non-serializable fields before serialing, and then setting the fields again to cause no side-effects to the object. 

This could also be done using transient, but since there are other types of serialization (Gson, Realm) in the project I thought this approach is less likely to break anything. 
If using transient is acceptable, I'll update my pull request.

Signed-off-by: Martin Kamleithner martin@diagnosia.com

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Improvements:
  -
 
 Bugfix:
- -  do not attempt to serialize unserializable fields on MatrixError in Event::writeExternal
+ -  remove unserializable fields in MatrixError
 
 API Change:
  -
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Improvements:
  -
 
 Bugfix:
- -
+ -  do not attempt to serialize unserializable fields on MatrixError in Event::writeExternal
 
 API Change:
  -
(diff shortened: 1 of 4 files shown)


bmarty:
Hello @knaeckeKami ,
Thanks for this PR.
Actually I do not know why `mErrorBody` and `mErrorBodyMimeType` has been added to `MatrixError` class, and they are not used at all.
I think the best solution is to remove them.
Benoît

bmarty:
@knaeckeKami will you be able to remove `mErrorBody` and `mErrorBodyMimeType`?

knaeckeKami:
Yes. I wasn't able to look into by now, but if it really isn't used anywhere I'll update my pull request today or tomorrow.

bmarty:
Thanks @knaeckeKami !

PR #418, opened 2019-01-23
MXCrypto: Add key backup passphrase support

bmarty:
Fixes vector-im/riot-android#2771

@uhoreg I'll be pleased to have some feedback from you, regarding the PBKDF2 algorithm implementation.

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/keysbackup/KeysBackupPassword.kt
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/keysbackup/KeysBackupPassword.kt
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2019 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


bmarty:
Thanks @uhoreg ,
A test has been added to check that it's producing the same result as riot web
Benoît

matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/keysbackup/KeysBackupPassword.kt
```diff
@@ -0,0 +1,135 @@
...
+    // noticed as dklen/hlen
+
+    // dklen = 256
+    // hlen = 512
+    val prf = Mac.getInstance("HmacSHA512")
+
+    prf.init(SecretKeySpec(password.toByteArray(), "HmacSHA512"))
+
+    // 256 bits key length
+    val dk = ByteArray(32)
```
manuroe: Could we use `OLMPkDecryption.privateKeyLength()` instead of magic 32?

PR #420, opened 2019-01-26
Fix RestClient exception in case of non-ASCII application label

### LINKED ISSUE #419: Unicode application label causes okhttp error in RestClient
The application label is used in creating `RestClient` user agent string. If the application label contains non-ASCII characters, which may be the case in localized apps, the `okhttp` library raises `IllegalArgumentException: Unexpected char ... at 0 in User-Agent value` exception.

I recommend checking application label and using application ID instead of label if there is any non-ASCII characters in it. I can submit a pull request if this is ok.

amiraliakbari:
Closes  #419 

### Pull Request Checklist

* [ ] Pull request is based on the develop branch
* [ ] Pull request updates [CHANGES.rst](https://github.com/matrix-org/matrix-android-sdk/blob/develop/CHANGES.rst)
* [ ] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Improvements:
  - MXCrypto: Add key backup passphrase support (vector-im/riot-android#2771).
 
 Bugfix:
- -
+ - Fix RestClient exception in case of non-ASCII application label (#419)
 
 API Change:
  -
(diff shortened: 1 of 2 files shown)


bmarty:
Hi @amiraliakbari ,

Thanks for the PR, I will have a look. I will be more confident if they where unit test to cover this, but they do not exist yet...

Benoît

bmarty:
Thanks, it's merged

PR #421, opened 2019-01-28
MXCrypto: ensure listeners are called on the UiThread

bmarty:


matrix-sdk/src/main/java/org/matrix/androidsdk/crypto/keysbackup/KeysBackup.kt
```diff
@@ -601,61 +615,63 @@ class KeysBackup(private val mCrypto: MXCrypto, session: MXSession) {
...
-                            Log.d(LOG_TAG, "checkAndStartKeyBackup: No usable key backup. version: " + keyBackupVersion.version)
+                            Log.d(LOG_TAG, "checkAndStartKeysBackup: No usable key backup. version: " + keyBackupVersion.version)
                             if (mKeysBackupVersion == null) {
                                 Log.d(LOG_TAG, "   -> not enabling key backup")
+                                mKeysBackupStateManager.state = KeysBackupStateManager.KeysBackupState.NotTrusted
                             } else {
                                 Log.d(LOG_TAG, "   -> disabling key backup")
-                                disableKeyBackup()
+                                disableKeysBackup()
+                                mKeysBackupStateManager.state = KeysBackupStateManager.KeysBackupState.Disabled
```
manuroe: Should be `NotTrusted` here too: There is backup on the homeserver but it is not usable.

bmarty: ok

PR #431, opened 2019-02-19
CreateRoomParams updates

giomfo:
- Bug Fix - setHistoryVisibility: remove existing value if any.
- Improvement: add `powerLevelContentOverride` param to override the default power level event.

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request updates [CHANGES.rst](https://github.com/matrix-org/matrix-android-sdk/blob/develop/CHANGES.rst)
* [ ] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/CreateRoomParams.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/CreateRoomParams.java
@@ -118,7 +118,7 @@ public class CreateRoomParams {
      * The power level content to override in the default power level event
      */
     @SerializedName("power_level_content_override")
-    public Map<String, Integer> powerLevelContentOverride;
+    public Map<String, Object> powerLevelContentOverride;
 
     /**
      * Add the crypto algorithm to the room creation parameters.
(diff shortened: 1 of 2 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/CreateRoomParams.java
```diff
@@ -114,6 +114,12 @@
     @SerializedName("is_direct")
     public Boolean isDirect;
 
+    /**
+     * The power level content to override in the default power level event
+     */
+    @SerializedName("power_level_content_override")
+    public Map<String, Integer> powerLevelContentOverride;
```
bmarty: Why not using type `PowerLevels` instead of a Map?

giomfo: @bmarty you are right but this would need to change the current PowerLevels class (to not be impacted by the default values)
FYI I updated the powerLevelContentOverride type to use the correct one

PR #438, opened 2019-02-26
Fix  / Room v3 permalinks don't always work #8315

BillCarsonFr:
Event v3 can contain +, so need to escape it when creating the permalink
+ On resolve permalink use the encoded uri fragment then split then decode

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@ Improvements:
 
 Bugfix:
  - Failed to send a video captured by the native camera. Replace the file scheme "file://" with "file:/" used by some Android devices.
+ - Fix / Escape room v3 event ids in permalinks (vector-im/riot-android#2981)
 
 API Change:
  -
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -9,6 +9,7 @@ Improvements:
 
 Bugfix:
  - Failed to send a video captured by the native camera. Replace the file scheme "file://" with "file:/" used by some Android devices.
+ - Fix / Room v3 permalinks don't always work #8315
```
bmarty: Use special link because this is a riot issue (also in commit message - force push to update it)

matrix-sdk/src/main/java/org/matrix/androidsdk/util/PermalinkUtils.java
```diff
@@ -109,12 +110,13 @@ public static String getLinkedId(@Nullable String url) {
...
     /**
      * Escape '/' in id, because it is used as a separator
+     * Escape '+' in room v3
      *
      * @param id the id to escape
      * @return the escaped id
      */
     private static String escape(String id) {
-        return id.replaceAll("/", "%2F");
+        return id.replaceAll("/", "%2F").replaceAll("\\+","%2B");
```
bmarty: Please format file :)

PR #440, opened 2019-03-04
Handle well-known data in the login response (vector-im/riot-android#3002)

bmarty:


matrix-sdk/src/main/java/org/matrix/androidsdk/HomeServerConnectionConfig.java
```diff
@@ -125,6 +127,37 @@ public Credentials getCredentials() {
      */
     public void setCredentials(Credentials credentials) {
         mCredentials = credentials;
+
+        // Override home server url and/or identity server url if provided
+        if (credentials.wellKnown != null) {
```
BillCarsonFr: why no &&?

PR #457, opened 2019-04-10
Fix crash on MXWebRtcView, reported by the PlayStore

bmarty:


bmarty:
Crash reproduced when cancelling an outgoing call (finish the call before it's established).
The PR fixes the issue

PR #474, opened 2019-07-10
F-Droid: sync once before suspend the events stream

giomfo:
F-Droid: sync once before suspend the events stream when a minimum delay is defined between 2 sync requests


CHANGES.rst
```diff
@@ -8,7 +8,7 @@ Improvements:
  -
 
 Bugfix:
- -
+ - F-Droid: sync once before suspend the events stream when a minimum delay is defined between 2 sync requests.
```
BillCarsonFr: Not sure to understand. what is it fixing?

giomfo: My plan was to apply a delay after the sync and not before. 
I thought the check on `mPendingDelayedIntent` [here](https://github.com/matrix-org/matrix-android-sdk/pull/474/files#diff-b76f857b869ad5b3392d8806fa3f91b3L533) was not a the right place. But now I'm less sure...

giomfo: Let's have a short sync on that?

giomfo:
After thinking, this change seems not required. I failed to reproduce the case where it would be useful...

PR #490, opened 2019-09-02
Support default stun server when no ice provided by HS

BillCarsonFr:
Support for default Stun server 

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request updates [CHANGES.rst](https://github.com/matrix-org/matrix-android-sdk/blob/develop/CHANGES.rst)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/call/MXCallsManager.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/call/MXCallsManager.java
@@ -93,6 +93,8 @@ public class MXCallsManager {
     // UI handler
     private final Handler mUIThreadHandler;
 
+    public static String defaultStunServerUri;
+
     /**
      * To create an outgoing call
      * 1- CallsManager.createCallInRoom()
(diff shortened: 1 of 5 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/call/MXCallsManager.java
```diff
@@ -223,12 +225,10 @@ private IMXCall createCall(String callId) {
         }*/
 
         // webrtc
-        if (null == call) {
-            try {
-                call = new MXWebRtcCall(mSession, mContext, getTurnServer());
-            } catch (Exception e) {
-                Log.e(LOG_TAG, "createCall " + e.getMessage(), e);
-            }
+        try {
```
bmarty: I assume the code above will never be uncommented, else we should keep the `if` test.

PR #491, opened 2019-09-05
Display correctly the revoked third-party invites.

giomfo:

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/core/EventDisplay.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/core/EventDisplay.java
@@ -297,7 +297,17 @@ public class EventDisplay {
                     displayName = redactedInfo;
                 }
 
-                text = mContext.getString(R.string.notice_room_third_party_invite, userDisplayName, displayName);
+                if (displayName != null) {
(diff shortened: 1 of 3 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/core/EventDisplay.java
```diff
@@ -297,7 +297,17 @@ public CharSequence getTextualDisplay(Integer displayNameColor, Event event, Roo
...
                 }
 
-                text = mContext.getString(R.string.notice_room_third_party_invite, userDisplayName, displayName);
+                if (displayName != null) {
+                    text = mContext.getString(R.string.notice_room_third_party_invite, userDisplayName, displayName);
+                } else {
+                    // Consider the invite has been revoked
+                    JsonObject prevContent = event.getPrevContentAsJsonObject();
+                    if (prevContent != null) {
+                        text = mContext.getString(R.string.notice_room_third_party_revoked_invite, userDisplayName, prevContent.get("display_name"));
```
bmarty: Handle the case where the display name is null?

giomfo: no this is not relevant...

PR #511, opened 2019-11-15
Fix #3382

BillCarsonFr:
Match identity server registration to the IS r0.3.0 spec
Fixes #3382


matrix-sdk/src/main/java/org/matrix/androidsdk/rest/api/IdentityAuthApi.kt
```diff
@@ -42,7 +42,7 @@ interface IdentityAuthApi {
      * @param body the body content
      */
     @POST("account/register")
-    fun register(@Body requestOpenIdTokenResponse: RequestOpenIdTokenResponse): Call<IdentityServerRegisterResponse>
+    fun register(@Body registerParamsParams: RegisterOpenIdTokenParams): Call<IdentityServerRegisterResponse>
```
bmarty: ParamsParams?

PR #512, opened 2019-11-18
Update from Weblate

RiotTranslateBot:


bmarty:
@bmarty There is a mistake, please check it after merging...

bmarty:
I reject because the new trad is buggy and I do not understand why the other commits come with thisPR

PR #548, opened 2020-05-29
Crypto: Only create one olm session at a time per device (vector-im/riot-android#3056). WIP

bmarty:
The PR is still WIP

@giomfo next step for me is to write an integration test, but maybe you can start to review what I have done and maybe confirm that the code fixes your problem.

### DIFF
--- a/matrix-sdk-crypto/src/main/java/org/matrix/androidsdk/crypto/data/MXOlmSessionResult.java
+++ b/matrix-sdk-crypto/src/main/java/org/matrix/androidsdk/crypto/data/MXOlmSessionResult.java
@@ -16,9 +16,7 @@
 
 package org.matrix.androidsdk.crypto.data;
 
-import java.io.Serializable;
-
-public class MXOlmSessionResult implements Serializable {
+public class MXOlmSessionResult {
     /**
      * the device
      */
@@ -30,6 +28,9 @@ public class MXOlmSessionResult implements Serializable {
      */
     public String mSessionId;
 
+    // True if mSessionId has been retrieved, (even if mSessionId is null)
+    public Boolean hasResult;
+
     /**
      * Constructor
      *
@@ -39,5 +40,6 @@ public class MXOlmSessionResult implements Serializable {
     public MXOlmSessionResult(MXDeviceInfo device, String sessionId) {
         mDevice = device;
         mSessionId = sessionId;
+        hasResult = sessionId != null && !sessionId.isEmpty();
     }
 }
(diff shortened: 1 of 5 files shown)


matrix-sdk-crypto/src/main/java/org/matrix/androidsdk/crypto/data/MXOlmSessionResult.java
```diff
@@ -30,6 +28,9 @@
      */
     public String mSessionId;
 
+    // True if mSessionId has been retrieved, (even if mSessionId is null)
```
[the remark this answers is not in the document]
bmarty: Yeap

matrix-sdk-crypto/src/main/java/org/matrix/androidsdk/crypto/internal/otk/OneTimeKeysResponseHandler.kt
```diff
@@ -116,4 +117,50 @@ class OneTimeKeysResponseHandler(
...
+        onError(usersDevicesToClaim) {
+            it.callback.onUnexpectedError(e)
+        }
+    }
+
+    private fun onError(usersDevicesToClaim: MXUsersDevicesMap<String>, callCallback: ((PendingRequest) -> Unit)) {
+        // Spread the failure to all the pending requests which are waiting for this session
+        val concernedPendingRequests = mutableSetOf<PendingRequest>()
+
+        pendingRequests.toList().forEach { pendingRequest ->
```
giomfo, line 143: Are you sur `mainForEach@` is not after `pendingRequests.toList().forEach` (line 143 instead of 144)?

bmarty: We have 3 nested `forEach`, that's why I used a label `mainForEach` for the second one, to be able to return form it when a `pendingRequest` is found in the `pendingRequests` Collection.
But we still continue to iterate on `pendingRequests`.

bmarty: Maybe the name with `main` is confusing then...

giomfo: sorry, but after thinking, I still think we should go to the next pendingRequest by jumping to the first forEach. I will need to sync irl

PR #552, opened 2020-06-25
Bug Fix - Killed Application misses some notifications

giomfo:
When the application is killed the first push notification is never displayed.
This was due to a bug in the RoomAccountData definition. Indeed some items of this class were not Serializable.
This make fail the loading of RoomAccountData from the store on new application launch
-> The store is considered corrupted -> the stream token is reseted
The first event which awakes the app is handled during an initial sync, so no push is applied for it...

### DIFF
--- a/matrix-sdk/src/main/java/org/matrix/androidsdk/data/RoomAccountData.java
+++ b/matrix-sdk/src/main/java/org/matrix/androidsdk/data/RoomAccountData.java
@@ -54,8 +54,8 @@ public class RoomAccountData implements java.io.Serializable {
     // The events the user wants to hide in this room.
     private Map<String, TaggedEventInfo> hiddenEvents = null;
 
-    // Store the content of all the provided events by using their event type.
-    private Map<String, JsonObject> eventContentsMap = new HashMap<>();
+    // Store all the provided events by using their event type.
+    private Map<String, Event> eventsMap = new HashMap<>();
 
     /**
      * Process an event that modifies room account data (like m.tag event).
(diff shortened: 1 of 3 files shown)


matrix-sdk/src/main/java/org/matrix/androidsdk/data/RoomAccountData.java
```diff
@@ -54,8 +54,8 @@
     // The events the user wants to hide in this room.
     private Map<String, TaggedEventInfo> hiddenEvents = null;
 
-    // Store the content of all the provided events by using their event type.
-    private Map<String, JsonObject> eventContentsMap = new HashMap<>();
+    // Store all the provided events by using their event type.
+    private Map<String, Event> eventsMap = new HashMap<>();
```
bmarty, line 58: Does an upgrade of the application with this change will work? I mean when the SDK will deserialize RoomAccountData, maybe this will cause issue.

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/TaggedEvents.kt
```diff
@@ -75,7 +76,7 @@ data class TaggedEventInfo(@JvmField
                            @JvmField
                            @SerializedName("tagged_at")
                            var taggedAt: Long? = null
-) {
+) : Serializable {
```
bmarty, line 79: Maybe `TaggedEventsContent` (in the same file)  should also implement `Serializable`?

matrix-sdk/src/main/java/org/matrix/androidsdk/data/RoomAccountData.java
giomfo: I consider the problem is already here. Currently the deserialisation failed on RoomAccountData, which is the bug I want to fix here. But you're right I will test again an update before merging

matrix-sdk/src/main/java/org/matrix/androidsdk/rest/model/TaggedEvents.kt
giomfo: Ok, this is not required for the moment, but this may be useful in the future. I will update this

giomfo: done

matrix-sdk/src/main/java/org/matrix/androidsdk/data/RoomAccountData.java
bmarty: ok

PR #553, opened 2020-06-29
Fix java.lang.NullPointerException:

giomfo:
at org.matrix.androidsdk.data.RoomAccountData.handleEvent (RoomAccountData.java:83)


giomfo:
Yes I wrapped all the block

PR #555, opened 2020-09-08
Fix a crash on the crypto code (race condition?)

bmarty:


matrix-sdk-crypto/src/main/java/org/matrix/androidsdk/crypto/verification/VerificationManager.kt
```diff
@@ -50,7 +54,7 @@ class VerificationManager(val session: CryptoSession) : VerificationTransaction.
 
     // Event received from the sync
     fun onToDeviceEvent(event: CryptoEvent) {
-        session.requireCrypto().getDecryptingThreadHandler().post {
+        mxCrypto.getDecryptingThreadHandler().post {
```
bmarty, line 57: We had a crash here.

giomfo (approved):
LGTM 
FYI I tested the change with success

PR #558, opened 2020-09-22
Set user id before storing it

yostyle:
Fix a crash when it checks user preference (related to matrix-org/synapse#7606)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Improvements:
  -
 
 Bugfix:
- - Fix a crash when it checks user preference (related to matrix-org/synapse#7606).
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -8,7 +8,7 @@ Improvements:
  -
 
 Bugfix:
- -
+ - Fix a crash when it checks user preference (related to matrix-org/synapse#7606).
```
bmarty: `preference`, did you mean `presence`?

bmarty (approved):
one remark, else LGTM

CHANGES.rst
yostyle: you right, I will update this.

PR #561, opened 2020-11-09
Fix crashes related to RoomAccountData

giomfo:
Flush the store if this isn't done yet to take into account the recent added eventsMap


bmarty (approved):
LGTM, thanks

PR #563, opened 2020-11-16
Improve the push rules handling

giomfo:
- Consider only the push rules data retrieved thanks to the server sync response
- Initialize the BingRulesManager with the stored push rules, instead of request them to the server
- Do not use anymore the C-S request to get the push rules directly (This is useless)
- Remove "onBingRulesUpdateListener" from BingRulesManager, it was not used

API Change:
- MXDataHandler: refreshPushRules() has been removed - we trust the server sync response
- BingRulesManager: loadRules() has been removed - we trust the server sync response
- BingRulesManager: an update of the push rules is not forced anymore when a rule is deleted, added or updated.
We let the server sync handle this push rules update.
- BingRulesManager: the interface onBingRulesUpdateListener has been removed. Use MXEventListener instead.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,12 +6,17 @@ Features:
 
 Improvements:
  - MXWebRtcCall: improve the mIsSupported flag handling.
+ - Improve/Simplify the push rules handling.
 
 Bugfix:
  - Fix a crash when it checks user presence (related to matrix-org/synapse#7606).
 
 API Change:
- -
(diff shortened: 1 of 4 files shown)


bmarty (approved):
LGTM, I like those type of PRs: 
image

But why you do not want to merge it on /develop?

yostyle (approved):

giomfo:
> But why you do not want to merge it on /develop?

I just wasn't sure API breaks were welcome on /develop

PR #567, opened 2020-12-09
Fix Crash: java.lang.NullPointerException: at org.matrix.androidsdk.data.RoomState.getMembersAsync

giomfo:
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void org.matrix.androidsdk.MXDataHandler.getMembersAsync(java.lang.String, org.matrix.androidsdk.core.callback.ApiCallback)' on a null object reference
at org.matrix.androidsdk.data.RoomState.getMembersAsync(RoomState.java:299)

see https://github.com/dinsic-pim/tchap-android/issues/680

This crash seems due to an unexpected race condition


bmarty (approved):
The fix looks like a workaround, but if it is OK for you it is OK for me

